PR #9, opened 2020-05-24
feat: adds no-multiple-stores rule

jsaguet:


src/rules/no-multiple-stores.ts
```diff
@@ -0,0 +1,56 @@
...
+      extraDescription: ['There is only one global store'],
+      recommended: 'error',
+    },
+    schema: [],
+    messages: {
+      [messageId]: 'Store should be injected only once',
+    },
+  },
+  defaultOptions: [],
+  create: context => {
```
timdeschryver, line 31: I wonder if we could do this by using just one query, something like this.
What do you think?

```ts
  create: context => {
    const constructors: TSESTree.Identifier[] = []

    return {
      [`MethodDefinition[kind='constructor'] Identifier>TSTypeAnnotation>TSTypeReference[typeName.name="Store"]`](
        node: TSESTree.TSTypeReference,
      ) {
        constructors.push(node.parent.parent as TSESTree.Identifier)
      },
      'Program:exit'() {
        constructors
          .filter((_, i) => i > 0)
          .forEach(node => {
            context.report({
              node,
              messageId,
            })
          })
      },
    }
  },
```

jsaguet: The issue I could see is that with the @typescript-eslint/parser, a constructor parameter with a visibility modifier is one level deeper than a parameter without. (There is an additional node TSParameterProperty with a visibility modifier)

For the following code:
image
ASTExplorer gives:
image

So I'm not sure we can do everything with 1 query because the parents won't be se same.

I spent a lot of time trying to get one query but didn't succeed because of that.
I'll have a look to confirm this later today

jsaguet: Ok, after a second look, I understand what you did there ! It is indeed much cleaner
I think we should consider renaming constructors to injectedStores. What do you think?

I couldn't find any documentation about the Program:exit hook. Do you know any?

jsaguet: Would this code work if there are multiple classes in the same file?

timdeschryver: Yea, that's true that it doesn't work with multiple classes.
I think for now I would keep it as simple as possible, we can always add more logic later to it.
But as you mentioned, I also think it there won't be multiple classes in the same file.

👍 for the rename.

jsaguet: Ok, I'll switch back to your implementation then

jsaguet: I removed the multi classes management.
I kept the error report when we exit from the constructor instead of waiting for the end of the program.
I wonder if it has better performance. What do you think?

timdeschryver: That's awesome, thanks!

PR #12, opened 2020-05-26
feat: adds no-effect-decorator rule

jsaguet:
I put this one as an error whereas it is only a warning in ngrx-tslint-rule.
createEffect has been introduced 2 versions ago in NGRX (if I'm not mistaken) so I believe it is ok to put it as an error.
Users can still override the recommended config and put it as a warning if needed.

If you believe it should stay a warning, I can switch it back.

### DIFF
--- a/src/configs/recommended.json
+++ b/src/configs/recommended.json
@@ -7,6 +7,7 @@
     "ngrx/no-typed-store": "error",
     "ngrx/avoid-dispatching-multiple-actions-sequentially": "error",
     "ngrx/no-multiple-stores": "error",
-    "ngrx/no-dispatch-in-effects": "error"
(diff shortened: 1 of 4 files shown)


timdeschryver:
I'm going to wait and merge #11 first.
That way we have this PR to test it :)

timdeschryver:
I suppose we can also remove the no-effect-decorator-and-creator rule, if we see this rule an error now?

timdeschryver:
I think the ci doesn't kick in because this PR was created before it was set up...

jsaguet:
That's true. Although it could still be relevant for users who would have set this no-effect-decorator as a warning only or even removed it.

timdeschryver:
You made a valid point, let's leave it there as it's already been migrated.

PR #14, opened 2020-05-26
feat: adds use-selector-in-select rule

jsaguet:
This the migration of the TSLint rule `ngrx-selector-for-select`
I made a small name change that made more sense to me.
What do you think ?

### DIFF
--- a/src/rules/utils/selectors/index.ts
+++ b/src/rules/utils/selectors/index.ts
@@ -23,3 +23,8 @@ export const ngModuleDecorator = `ClassDeclaration > Decorator > CallExpression[
 export const ngModuleProviders = `${ngModuleDecorator} ObjectExpression Property[key.name='providers'] > ArrayExpression Identifier`
 
 export const ngModuleImports = `${ngModuleDecorator} ObjectExpression Property[key.name='imports'] > ArrayExpression CallExpression[callee.object.name='EffectsModule'][callee.property.name=/forRoot|forFeature/] ArrayExpression > Identifier`
+
+const pipeableSelect = `CallExpression[callee.property.name="pipe"] CallExpression[callee.name="select"]`
+const storeSelect = `CallExpression[callee.object.name='store'][callee.property.name='select']`
+
+export const select = `${pipeableSelect} Literal, ${storeSelect} Literal, ${pipeableSelect} ArrowFunctionExpression, ${storeSelect} ArrowFunctionExpression`
--- a/tests/rules/use-selector-in-select.test.ts
+++ b/tests/rules/use-selector-in-select.test.ts
@@ -1,0 +1,135 @@
+import { stripIndent } from 'common-tags'
+import rule, {
+  ruleName,
+  messageId,
+} from '../../src/rules/use-selector-in-select'
+import { ruleTester } from '../utils'
+
+ruleTester().run(ruleName, rule, {
+  valid: [
+    `store.pipe(select(selectCustomers))`,
+    `store.pipe(select(selectorsObj.selectCustomers))`,
+    `store.select(selectCustomers)`,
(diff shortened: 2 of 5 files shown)


tests/rules/use-selector-in-select.test.ts
```diff
@@ -0,0 +1,99 @@
...
+import rule, {
+  ruleName,
+  messageId,
+} from '../../src/rules/use-selector-in-select'
+import { ruleTester } from '../utils'
+
+ruleTester().run(ruleName, rule, {
+  valid: [
+    stripIndent`
+    const selectCustomers = createSelector(
```
timdeschryver, line 11: I think all the `createSelectors` can be removed and that it would be enough to write on valid tests as:

```ts
// 1
store.pipe(select(selectCustomers))

// 2
store.pipe(select(selectCustomers))
```

jsaguet: You're right, let's keep it simple

jsaguet:
As a side note, we are getting a lot of warning messages during tests:
image
Maybe we should update @typescript-eslint packages ?

timdeschryver:
Yea, I was updating some stuff yesterday but I didn't want to update too much stuff as it could perhaps have blocked your PRs 😅

src/rules/utils/selectors/index.ts
```diff
@@ -23,3 +23,6 @@ export const ngModuleDecorator = `ClassDeclaration > Decorator > CallExpression[
 export const ngModuleProviders = `${ngModuleDecorator} ObjectExpression Property[key.name='providers'] > ArrayExpression Identifier`
 
 export const ngModuleImports = `${ngModuleDecorator} ObjectExpression Property[key.name='imports'] > ArrayExpression CallExpression[callee.object.name='EffectsModule'][callee.property.name=/forRoot|forFeature/] ArrayExpression > Identifier`
+
+export const selectWithLiteral = `CallExpression[callee.property.name="pipe"] CallExpression[callee.name="select"] Literal`
```
timdeschryver, line 27: After a second look, we can create another query that we could reuse as a `select` query select 

```
CallExpression[callee.property.name="pipe"] CallExpression[callee.name="select"]
```

While I was think about this, I also thought that we should support `select` function on the `store`, e.g. `store.select(selectCustomers)`. Do you also want to add it to this PR?

jsaguet: Good idea!
I added support for store select and refactored a little to share eslint selectors

jsaguet: I can even make one selector for everything actually

src/rules/utils/selectors/index.ts
```diff
@@ -23,3 +23,8 @@ export const ngModuleDecorator = `ClassDeclaration > Decorator > CallExpression[
 export const ngModuleProviders = `${ngModuleDecorator} ObjectExpression Property[key.name='providers'] > ArrayExpression Identifier`
 
 export const ngModuleImports = `${ngModuleDecorator} ObjectExpression Property[key.name='imports'] > ArrayExpression CallExpression[callee.object.name='EffectsModule'][callee.property.name=/forRoot|forFeature/] ArrayExpression > Identifier`
+
+const pipeableSelect = `CallExpression[callee.property.name="pipe"] CallExpression[callee.name="select"]`
+const storeSelect = `CallExpression[callee.object.name='store'][callee.property.name='select']`
+
+export const select = `${pipeableSelect} Literal, ${storeSelect} Literal, ${pipeableSelect} ArrowFunctionExpression, ${storeSelect} ArrowFunctionExpression`
```
timdeschryver, line 30: TIL you can add multiple selectors with `,`. 🤯

PR #18, opened 2020-05-30
feat: adds on-reducer-explicit-return-type rule

jsaguet:
Here is the newly added rule.

I must say I am not a big fan of the rule name. "On reducer" makes me think we are talking about the reducer itself but that's not really the case.

I thought of something like "state-change-function-explicit-return-type" but it's a bit long and I don't know if everybody calls these functions "state change function". I found this name in NGRX documentation though.

"on-function-explicit-return-type" would be a bit clearer to me but we're still not typing the "on" function either.

As it was done on tslint repo, I didn't add the rule to the recommended config although I don't fully understand why we would not.

### DIFF
--- a/src/configs/recommended.json
+++ b/src/configs/recommended.json
@@ -9,6 +9,7 @@
     "ngrx/no-dispatch-in-effects": "error",
     "ngrx/no-effect-decorator": "error",
     "ngrx/no-effects-in-providers": "error",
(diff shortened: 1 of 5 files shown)


jsaguet:
Changed the rule name and added it to recommended config

timdeschryver:
@all-contributors please add @jsaguet for ideas
@all-contributors please add @StephenCooper for ideas

PR #22, opened 2020-05-30
test: verify all rules are exported

timdeschryver:


src/rules/avoid-dispatching-multiple-actions-sequentially.ts
```diff
@@ -16,9 +16,6 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
     docs: {
       category: 'Possible Errors',
       description: 'It is recommended to only dispatch one action at a time.',
-      extraDescription: [
-        'See more at https://redux.js.org/style-guide/style-guide#avoid-dispatching-many-actions-sequentially',
-      ],
```
jsaguet, line 21: Can't this be appended to the description item instead of being deleted? extraDescription often provides useful insights

timdeschryver: I was planning to revisit all the descriptions later (because I think some of them could be better worded), at the same time I also wanted to provide more info in the docs (introduced in #17). 

Do you think we should wait to merge this in, until this is done?

jsaguet: No it's ok, it will still be possible to find the text if needed when the doc will be written

PR #23, opened 2020-05-30
fix: do not report on-function-explicit-return-type error when using immer

jsaguet:
The selector was not correct and reported an error when using immer produce function.

### DIFF
--- a/tests/rules/on-function-explicit-return-type.test.ts
+++ b/tests/rules/on-function-explicit-return-type.test.ts
@@ -25,6 +25,16 @@ ruleTester().run(ruleName, rule, {
       on(increment, s => incrementFunc(s)),
       on(increment, (s): State => incrementFunc(s)),
     )`,
+    `
+    const reducer = createReducer(
+      initialState,
+      on(
+        increment,
+        produce((draft: State, action) => {
+            draft.counter++;
+        }),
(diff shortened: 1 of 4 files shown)


tests/rules/on-function-explicit-return-type.test.ts
```diff
@@ -24,6 +24,12 @@ ruleTester().run(ruleName, rule, {
       on(increment, incrementFunc),
       on(increment, s => incrementFunc(s)),
       on(increment, (s): State => incrementFunc(s)),
+      on(
```
timdeschryver, line 27: Can this be a separate test please?

jsaguet: done

jsaguet:
Which behavior do you expect with ngrx-etc? Error or pass?
Currently, if I'm not mistaken, it should pass since you're either using `mutableOn` or `createMutableReducer` and the rule looks for both `createReducer`AND `on`

timdeschryver:
It should pass.
🤦‍♂️ you're right, which is great as we don't have to do anything for it 🎉

For more context: It's in the TSLint repo where some rules are looking for `on` and `mutableOn`...
Thanks! LGTM

PR #24, opened 2020-05-30
feat: add no-reducer-in-key-names rule

jsaguet:


src/rules/no-reducer-in-key-names.ts
```diff
@@ -0,0 +1,57 @@
...
+      return name.toLowerCase().includes(excludedKeyword)
+    }
+
+    return {
+      [`${storeActionReducerMap}, ${actionReducerMap}`](
+        node: TSESTree.Property,
+      ) {
+        const key = node.key
+        if (
+          (key.type === AST_NODE_TYPES.Literal &&
```
timdeschryver, line 45: I would prefer this check implemented as a guard (similar to https://github.com/timdeschryver/eslint-plugin-ngrx/blob/v1.0.0/src/rules/utils/helper-functions/index.ts#L3)
It will have as result that the node will have the correct type, thus `key.value as string` would not be needed normally.

Thoughts?

jsaguet: I implemented the guard because it's more readable. However key.value on a literal can be many things so `key.value as string` is still necessary
image

jsaguet: It seems I can use raw property which is already a string. So that's what I've done

timdeschryver: 👍

jsaguet:
fixed conflict

PR #25, opened 2020-05-31
feat: adds no-multiple-actions-in-effects rule

jsaguet:
This rule's implementation gave me headaches... Working with types in ESLint is not that easy after all.

I didn't manage to check if the return is of type `Action[]` so I've only been looking for returned arrays in effects.
It covers all test cases that I found in ngrx-tslint-rules but I am not sure it will not report false positives.

What do you think?

### DIFF
--- a/tsconfig.eslint.json
+++ b/tsconfig.eslint.json
@@ -1,4 +1,5 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["**/*.ts"]
+  "include": ["**/*.ts"],
+  "files": ["fake.ts"]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -8955,6 +8955,12 @@
         "tslib": "^1.8.1"
       }
     },
+    "tsutils-etc": {
+      "version": "1.2.2",
+      "resolved": "https://registry.npmjs.org/tsutils-etc/-/tsutils-etc-1.2.2.tgz",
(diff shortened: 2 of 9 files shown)


timdeschryver:
I will get back to this PR the next days.
I'm currently a bit busy.
Thanks in advance!

tsconfig.eslint.json
```diff
@@ -1,4 +1,5 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["**/*.ts"]
+  "include": ["**/*.ts"],
+  "files": ["fake.ts"]
```
timdeschryver, line 4: do we need this?

jsaguet: Yes, otherwise an error is thrown when running tests:
` A fatal parsing error occurred: Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
    The file does not match your project config: fake.ts.
    The file must be included in at least one of the projects provided.`

timdeschryver: 👍
Weird I would've expected that the file should exist.

timdeschryver: @jsaguet could we create an eslint config in the test folder and use that on in our tests? It can be a simple config, something simalar to the [rxjs eslint plugin](https://github.com/cartant/eslint-plugin-rxjs/tree/master/tests)

jsaguet: I don't understand why, but I didn't manage to make it work with such a file

timdeschryver:
I'm going to merge this in, and try something later.

PR #28, opened 2020-06-03
feat: add select-style rule

jsaguet:
Add a rule to define how selectors should be used in the project.

This rule is not part of recommended configuration.

The default value is 'operator' because it seems to be the recommended way in NGRX documentation and it allows to use more advanced features: 
- [Make use of rxjs operators](https://ngrx.io/guide/store/selectors#select-a-non-empty-state-using-pipeable-operators)
- [Extracting a pipeable operator](https://ngrx.io/guide/store/selectors#solution-extracting-a-pipeable-operator)

For ease (and because I use the underlying modifications), I'm including in this branch the commit for #27. So the bug fix should probably be merged before this one

### DIFF
--- a/src/rules/select-style.ts
+++ b/src/rules/select-style.ts
@@ -1,0 +1,67 @@
+import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
+
+import { docsUrl, pipeableSelect, storeSelect } from '../utils'
+
+export const ruleName = 'select-style'
+
+export const methodSelectMessageId = 'storeSelect'
+export const operatorSelectMessageId = 'operatorSelect'
+
+export type MessageIds =
+  | typeof methodSelectMessageId
+  | typeof operatorSelectMessageId
+
+export const OPERATOR = 'operator'
+export const METHOD = 'method'
+
+type Options = [{ mode: string }]
+
+export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
+  name: ruleName,
+  meta: {
+    type: 'problem',
+    docs: {
+      category: 'Possible Errors',
+      description: `Selectors can be used either with 'select' as a pipeable operator or as a method`,
+      recommended: false,
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          mode: {
+            enum: [OPERATOR, METHOD],
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+    messages: {
+      [methodSelectMessageId]:
+        'Selectors should be used with select method: this.store.select(selector)',
+      [operatorSelectMessageId]:
+        'Selectors should be used with the pipeable operator: this.store.pipe(select(selector))',
+    },
+  },
+  defaultOptions: [{ mode: OPERATOR }],
+  create: (context, [{ mode }]) => {
+    return {
+      [pipeableSelect](node: TSESTree.CallExpression) {
+        if (mode === METHOD) {
+          context.report({
+            node,
+            messageId: methodSelectMessageId,
+          })
+        }
+      },
+      [storeSelect](node: TSESTree.Identifier) {
+        if (mode === OPERATOR) {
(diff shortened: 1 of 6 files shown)


src/rules/select-style.ts
```diff
@@ -0,0 +1,62 @@
...
+    messages: {
+      [methodSelectMessageId]:
+        'Selectors should be used with select method: this.store.select(selector)',
+      [operatorSelectMessageId]:
+        'Selectors should be used with the pipeable operator: this.store.pipe(select(selector))',
+    },
+  },
+  defaultOptions: [OPERATOR],
+  create: (context) => {
+    const mode: string = context.options[0] || OPERATOR
```
timdeschryver, line 42: Is it needed to fallback to `OPERATOR`? I thought this would  be handled by the default options config.
Perhaps for this to work, we should use the following syntax (taken from [angular eslint](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/rules/component-class-suffix.ts#L48)).

```suggestion
  create(context, [options]) {
    const { mode } = options;
```

src/rules/select-style.ts
```diff
@@ -0,0 +1,62 @@
...
+export const operatorSelectMessageId = 'operatorSelect'
+
+export type MessageIds =
+  | typeof methodSelectMessageId
+  | typeof operatorSelectMessageId
+
+export const OPERATOR = 'operator'
+export const METHOD = 'method'
+
+type Options = [string]
```
timdeschryver, line 17: WDYT about the following syntax for it? I've mostly seen it this way

```suggestion
type Options = [{
  mode: string
}]
```

src/rules/select-style.ts
```diff
@@ -0,0 +1,62 @@
...
+      [storeSelect](node: TSESTree.Identifier) {
+        if (mode === OPERATOR) {
+          context.report({
+            node,
+            messageId: operatorSelectMessageId,
+          })
+        }
+      },
+    }
+  },
```
timdeschryver, line 61: I wonder if the following would be faster and if it's worth it:
If I got the syntax right it will only add one selector.

```suggestion
 return {
      ...(mode === METHOD ? {[pipeableSelect](node: TSESTree.CallExpression) {
        context.report({
          node,
          messageId: methodSelectMessageId,
        })
      }} : {}),
      ...(mode === OPERATOR ? {[storeSelect](node: TSESTree.Identifier) {
        context.report({
          node,
          messageId: operatorSelectMessageId,
        })
      }} : {}),
}
```

jsaguet: It would probably be a bit faster but it's much harder to read imho. 
I think it could also be more difficult to extend it by adding more options if required in the future.
All in all, I am not sure the performance gain is worth it.

We could also write it like this (which I find more readable):
```suggestion
create: (context, [{ mode }]) => {
    if (mode === METHOD) {
      return {
        [pipeableSelect](node: TSESTree.CallExpression) {
          context.report({
            node,
            messageId: methodSelectMessageId,
          })
        },
      }
    }

    if (mode === OPERATOR) {
      return {
        [storeSelect](node: TSESTree.Identifier) {
          context.report({
            node,
            messageId: operatorSelectMessageId,
          })
        },
      }
    }
  },
```
But I never saw any rule written like this in the few I have read so I guess it's not really in the ESLint spirit.

jsaguet:
> 
> 
> This is looking good!
> I left a few comments/questions.
> 
> If it's OK for you, we can close #27 and merge this PR in.

As you wish! I don't have any opinion on this 😄

src/rules/select-style.ts
jsaguet: I am really not sure the performance difference would be noticeable.
Do you know any way we can do performance tests to get some data?
(if it's worth investing time in this)

timdeschryver:
I don't think it's worth it tbh. Thanks for providing your feedback and
keeping it simple

PR #35, opened 2020-06-21
refactor: create recommended config based on rules config

timdeschryver:
Thoughts @jsaguet ?


jsaguet:
Hi @timdeschryver,
I think it's great to automate the recommended config creation but I am not sure about this implementation.

It looks like this change would make eslint "calculate" all recommended rules at runtime.

Does this work when the plugin is integrated in a project ?
If yes, it probably has some performance cost for end users for every time eslint is run on one file or their entire project.

I've seen @typescript-eslint package has [a script in their package.json](https://github.com/typescript-eslint/typescript-eslint/blob/26da8de7fcde9eddec63212d79af781c4bb22991/packages/eslint-plugin/package.json#L38) to generate the file.
The script itself is basically a more complicated version of your implementation. [Available here](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/tools/generate-configs.ts)

I believe you can have multiple advantages:
- You can run it manually or on pre-commit hook for example. The test on the recommended config would catch if the script was not run.
- It adds no further computation time for end users.
- It's more readable for anyone looking through the repository to understand which rules are included in the recommended one.
- You have better separation of concerns: the recommended file lists the rule (as it's supposed to) and your automation task is in a separate file.

What do you think?

timdeschryver:
@jsaguet that's great feedback, thanks!
I do think that performance wise it doesn't make a difference, it's just an iteration over a small object.
I agree that it will be more clear and clean if we move it to a hook! I will look into that later.

scripts/generate-config.ts
```diff
@@ -0,0 +1,30 @@
+import fs from 'fs'
+import path from 'path'
```
jsaguet, line 2: you can even import only the functions you need
```
import { writeFileSync } from 'fs';
import { join } from 'path';
```

PR #46, opened 2020-07-22
fix(use-selector-in-select): allow parameters

### LINKED ISSUE #41: Bug: Incorrect error report for use-selector-in-select
I noticed that the rule `use-selector-in-select` incorrectly reports the following error `using string or props drilling is not prefered, use a selector instead` for this code:

`this.store.pipe(select(selectQueryParam('parameter'))`

`selectQueryParam` is a selector from `@ngrx/router-store` and `parameter` is a parameter of the selector.

timdeschryver:
Closes #41

@jsaguet this is basically a re-write of the rule.
The rule also takes into account when the store is renamed to for example `store$`

### DIFF
--- a/tests/rules/no-multiple-stores.test.ts
+++ b/tests/rules/no-multiple-stores.test.ts
@@ -26,32 +26,36 @@ ruleTester().run(ruleName, rule, {
   invalid: [
     fromFixture(
       stripIndent`
-      export class NotOkNoVisibility {
-        constructor(store: Store, store2: Store){}
-                                  ~~~~~~~~~~~~~ [${messageId}]
-      }`,
+        export class NotOkNoVisibility {
+          constructor(store: Store, store2: Store){}
+                      ~~~~~~~~~~~~                  [${messageId}]
+                                    ~~~~~~~~~~~~~   [${messageId}]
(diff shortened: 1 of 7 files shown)


tests/rules/no-multiple-stores.test.ts
```diff
@@ -11,82 +11,63 @@ ruleTester().run(ruleName, rule, {
...
-        constructor(store: Store<{}>, store2: Store<{}>){}
+        constructor(store: Store, store2: Store){}
       }`,
       errors: [
         {
           messageId,
           line: 2,
-          column: 33,
-          endLine: 2,
-          endColumn: 50,
```
jsaguet, line 37: Any particular reason to remove these constraints ?

timdeschryver: Going to rewrite them with a helper function from eslint-etc.
I did the rest already, see https://github.com/timdeschryver/eslint-plugin-ngrx/commit/81a34734dcee5047df0ef56c888c9fbfe511eab3

jsaguet: Interesting !

PR #53, opened 2020-11-21
fix: update eslint-plugin-rxjs version

jsaguet:
version update to solve https://github.com/cartant/eslint-plugin-rxjs/issues/54 issue.

I was wondering if peer dependency wouldn't be better to let the user select the version to use.
What do you think ?


timdeschryver (approved):

timdeschryver:
Thanks @jsaguet 
About the peer dependency, how would this react when you don't have the rxjs plugin installed or when you're using a version in which a lint rule isn't included.

PR #54, opened 2020-11-22
fix: sync preferred select usage

timdeschryver:


jsaguet:
Hello @timdeschryver 
Is this change done to better reflect NGRX doc ?

I see that documentation refers mainly to the select method in v10 whereas it was the pipeable select in v9.
Do you know the reason ?

timdeschryver:
It's the preferred way.
Reasons are:

- better dev ergonomics (less to type and import)
- you can mock `store.select`

PR #62, opened 2021-01-19
fix: no-multiple-actions-in-effects not reporting mapTo operators

### LINKED ISSUE #61: [no-multiple-actions-in-effects] not reporting for *-MapTo operators
Example:

```ts
export class Effects {
  readonly test$ = createEffect(() =>
    this.actions$.pipe(ofType(Actions.test), switchMapTo([foo(), bar()])),
  );
}
```

It seems that `*MapTo` operators are missing from:

https://github.com/timdeschryver/eslint-plugin-ngrx/blob/48afe768e45da866681d59df0fe8229aeb7ddf5e/src/utils/selectors/index.ts#L31

rafaelss95:
Fixes #61.

in addition to fix the error reported in #61, I also changed the selectors to search for the exact names, instead of part of them, to avoid a possible problem with custom operators and also added some more tests to cover for all operators in the context.

### DIFF
--- a/src/utils/selectors/index.ts
+++ b/src/utils/selectors/index.ts
@@ -28,8 +28,13 @@ export const storeActionReducerMap = `${ngModuleDecorator} ObjectExpression Prop
 
 export const actionReducerMap = `VariableDeclarator[id.typeAnnotation.typeAnnotation.typeName.name='ActionReducerMap'] > ObjectExpression > Property`
 
-const effectsOperator = `ClassProperty > CallExpression[callee.name='createEffect'] CallExpression[callee.name=/switchMap|concatMap|mergeMap|flatMap|exhaustMap/]`
+const createEffectExpression = `ClassProperty > CallExpression[callee.name='createEffect']`
 
-export const effectsArrowReturn = `${effectsOperator} > ArrowFunctionExpression > ArrayExpression`
+const mapOperators = '(concat|exhaust|flat|merge|switch)Map'
+const mapToOperators = '(concat|merge|switch)MapTo'
+const mapOperatorsExpression = `${createEffectExpression} CallExpression[callee.name=/^${mapOperators}$/]`
+const mapToOperatorsExpression = `${createEffectExpression} CallExpression[callee.name=/^${mapToOperators}$/]`
 
-export const effectsReturn = `${effectsOperator} ReturnStatement`
+export const effectsImplicitReturn = `${mapOperatorsExpression} > ArrowFunctionExpression > ArrayExpression, ${mapToOperatorsExpression} ArrayExpression`
+
+export const effectsReturn = `${mapOperatorsExpression} ReturnStatement`
--- a/tests/rules/no-multiple-actions-in-effects.test.ts
+++ b/tests/rules/no-multiple-actions-in-effects.test.ts
@@ -37,6 +37,54 @@ ruleTester().run(ruleName, rule, {
     {
       code: `
         export class Effects {
+          thirteen$ = createEffect(() =>
+            this.actions$.pipe(mapTo(foo()))
+          )
+        }`,
+      parserOptions: {
+        project: './tsconfig.tests.eslint.json',
+      },
+      filename: 'test.ts',
+    },
+    {
+      code: `
+        export class Effects {
+          eleven$ = createEffect(() =>
+            this.actions$.pipe(switchMapTop([foo()]))
+          )
+        }`,
+      parserOptions: {
+        project: './tsconfig.tests.eslint.json',
+      },
+      filename: 'test.ts',
+    },
+    {
+      code: `
+        export class Effects {
+          twelve$ = createEffect(() =>
+            this.actions$.pipe(aconcatMapTo([bar()]))
+          )
+        }`,
+      parserOptions: {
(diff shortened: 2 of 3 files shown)


src/utils/selectors/index.ts
```diff
@@ -28,8 +28,13 @@ export const storeActionReducerMap = `${ngModuleDecorator} ObjectExpression Prop
...
+const createEffectExpression = `ClassProperty > CallExpression[callee.name='createEffect']`
 
-export const effectsArrowReturn = `${effectsOperator} > ArrowFunctionExpression > ArrayExpression`
+const mapOperators = '(concat|exhaust|flat|merge|switch)Map'
+const mapToOperators = '(concat|merge|switch)MapTo'
+const mapOperatorsExpression = `${createEffectExpression} CallExpression[callee.name=/^${mapOperators}$/]`
+const mapToOperatorsExpression = `${createEffectExpression} CallExpression[callee.name=/^${mapToOperators}$/]`
 
-export const effectsReturn = `${effectsOperator} ReturnStatement`
+export const effectsImplicitReturn = `${mapOperatorsExpression} > ArrowFunctionExpression > ArrayExpression, ${mapToOperatorsExpression} ArrayExpression`
```
rafaelss95, line 38: I'm not sure about all of these variable names. That's what came to mind this time, feel free to suggest :)

tests/rules/no-multiple-actions-in-effects.test.ts
```diff
@@ -34,6 +34,54 @@ ruleTester().run(ruleName, rule, {
...
+      parserOptions: {
+        project: './tsconfig.tests.eslint.json',
+      },
+      filename: 'test.ts',
+    },
+    {
+      code: `
+        export class Effects {
+          eleven$ = createEffect(() =>
+            this.actions$.pipe(switchMapTop([foo()]))
```
timdeschryver, line 53:
```suggestion
            this.actions$.pipe(switchMapTo([foo()]))
```

tests/rules/no-multiple-actions-in-effects.test.ts
```diff
@@ -34,6 +34,54 @@ ruleTester().run(ruleName, rule, {
...
+      parserOptions: {
+        project: './tsconfig.tests.eslint.json',
+      },
+      filename: 'test.ts',
+    },
+    {
+      code: `
+        export class Effects {
+          twelve$ = createEffect(() =>
+            this.actions$.pipe(aconcatMapTo([bar()]))
```
timdeschryver, line 65:
```suggestion
            this.actions$.pipe(concatMapTo([bar()]))
```

src/utils/selectors/index.ts
```diff
@@ -28,8 +28,13 @@ export const storeActionReducerMap = `${ngModuleDecorator} ObjectExpression Prop
 
 export const actionReducerMap = `VariableDeclarator[id.typeAnnotation.typeAnnotation.typeName.name='ActionReducerMap'] > ObjectExpression > Property`
 
-const effectsOperator = `ClassProperty > CallExpression[callee.name='createEffect'] CallExpression[callee.name=/switchMap|concatMap|mergeMap|flatMap|exhaustMap/]`
+const createEffectExpression = `ClassProperty > CallExpression[callee.name='createEffect']`
 
-export const effectsArrowReturn = `${effectsOperator} > ArrowFunctionExpression > ArrayExpression`
+const mapOperators = '(concat|exhaust|flat|merge|switch)Map'
+const mapToOperators = '(concat|merge|switch)MapTo'
```
timdeschryver, line 34: Nice!

timdeschryver (changes requested):
I spotted some typos, for the rest it's good - thanks!

timdeschryver:
Could we also add a test for an invalid `mapTo` ? 
I'm asking because I'm not 100% sure if we catch them with the used RegExp.

rafaelss95:
Hey @timdeschryver, about the _typos_:

Well, actually I did this intentionally :stuck_out_tongue_closed_eyes:.

I've described in the PR description:

> I also changed the selectors to search for the exact names, instead of part of them, to avoid a possible problem with custom operators and also added some more tests to cover for all operators in the context.

Let me know if you got it or if you think it's unnecessary so I can remove it.

> Could we also add a test for an invalid `mapTo` ?
> I'm asking because I'm not 100% sure if we catch them with the used RegExp.

Hmm, should this rule report `map` and `mapTo`? Shouldn't we only report the operators that return observable? Let me know so I can change it appropriately.

timdeschryver:
Woops 😅
To be honest, I don't really use the `mapTo` operator in effects, but I suppose there will be usages of `mapTo` that return multiple actions. If we add the `mapTo` operators, perhas we should also add the "normal" `mapTo` operator?

rafaelss95:
So, we can't return `mapTo` in a `effect`:

Captura de Tela 2021-01-20 às 09 49 43
Captura de Tela 2021-01-20 às 09 50 50

timdeschryver:
Then let's get this merged! Thanks a lot

timdeschryver (approved):

timdeschryver:
@all-contributors please add @rafaelss95  for code, tests

PR #67, opened 2021-01-31
feat: add fixer for use-lifecycle-interface

rafaelss95:
This PR aims to add a fixer for `use-lifecycle-interface`.

**Before**:
```ts
class UserEffects {
  ngrxOnInitEffects() {}
}
```

**After (--fix)**:
```ts
import { OnInitEffects } from '@ngrx/effects';
class UserEffects implements OnInitEffects {
  ngrxOnInitEffects() {}
}
```

### DIFF
--- a/src/rules/use-lifecycle-interface.ts
+++ b/src/rules/use-lifecycle-interface.ts
@@ -1,6 +1,14 @@
-import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
+import { ESLintUtils } from '@typescript-eslint/experimental-utils'
+import type { TSESTree } from '@typescript-eslint/experimental-utils'
 
-import { docsUrl, findClassDeclarationNode, isIdentifier } from '../utils'
+import {
+  docsUrl,
+  findClassDeclarationNode,
+  findImportDeclarationNode,
+  getConditionalImportFix,
+  getImplementsSchemaFixer,
+  isIdentifier,
+} from '../utils'
 
 export const ruleName = 'use-lifecycle-interface'
 
@@ -9,13 +17,13 @@ export type MessageIds = typeof messageId
 
 type Options = []
 
-const lifecycles = [
-  'ngrxOnInitEffects',
-  'ngrxOnRunEffects',
-  'ngrxOnIdentifyEffects',
-]
-
-const lifecyclesRegexp = lifecycles.join('|')
+const lifecycleMapper = {
+  ngrxOnInitEffects: 'OnInitEffects',
+  ngrxOnRunEffects: 'OnRunEffects',
+  ngrxOnIdentifyEffects: 'OnIdentifyEffects',
+} as const
+const effectsModulePath = '@ngrx/effects'
+const lifecyclesPattern = Object.keys(lifecycleMapper).join('|')
 
 export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   name: ruleName,
@@ -27,6 +35,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
         'Ensures classes implement lifecycle interfaces corresponding to the declared lifecycle methods',
       recommended: 'warn',
     },
+    fixable: 'code',
     schema: [],
     messages: {
       [messageId]:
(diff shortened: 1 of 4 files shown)


src/utils/helper-functions/utils.ts
```diff
@@ -10,3 +10,21 @@ export function findClassDeclarationNode(
...
+  while ((parentNode = parentNode.parent)) {
+    if (parentNode.type !== 'Program') continue
+
+    return parentNode.body.find(
+      (node) => isImportDeclaration(node) && node.source.value === module,
+    ) as TSESTree.ImportDeclaration | undefined
+  }
+
+  // TODO: Remove this when enable tsConfig strict option.
+  return parentNode as TSESTree.ImportDeclaration | undefined
```
rafaelss95, line 29: When I was implementing this, I noticed that `strict` is not enabled. I could send a follow-up PR if desired.

timdeschryver: Good catch! Feel free to enable strict mode

src/utils/helper-functions/guards.ts
```diff
@@ -31,3 +31,15 @@ export function isClassDeclaration(
 ): node is TSESTree.ClassDeclaration {
   return node && node.type === AST_NODE_TYPES.ClassDeclaration
 }
+
+export function isImportDeclaration(
+  node: TSESTree.Node,
+): node is TSESTree.ImportDeclaration {
+  return node.type === 'ImportDeclaration'
```
timdeschryver, line 38: we should also check if `node` isn't `undefined` so it behaves the same as the other guards.

src/utils/helper-functions/guards.ts
```diff
@@ -31,3 +31,15 @@ export function isClassDeclaration(
...
+}
+
+export function isImportSpecifier(
+  node: TSESTree.Node,
+): node is TSESTree.ImportSpecifier {
+  return node.type === 'ImportSpecifier'
```
timdeschryver, line 44: Same here

src/utils/helper-functions/guards.ts
timdeschryver, line 38: We should also check if `node` is `undefined`, this way it behaves like the other guards

src/utils/helper-functions/guards.ts
timdeschryver, line 44: Same here

src/rules/use-lifecycle-interface.ts
```diff
@@ -36,27 +45,44 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
-        const methodName = node.name
-        const interfaceName = methodName.substr(4)
-
-        if (!implementz.includes(interfaceName)) {
-          context.report({
-            node,
-            messageId,
-            data: {
-              methodName,
+        const {
```
timdeschryver, line 65: Can we move this to inside the fixer scope?
This way this only gets executed when someone wants to fix the the code.

timdeschryver (commented):
This is great, thanks @rafaelss95 !
I think this is our first rule with a fixer, do you think we should add a note to the docs?

src/utils/helper-functions/utils.ts
rafaelss95: Done, #70.

Btw, thanks for adding me as a collaborator. It's a pleasure to help in this project :)

src/utils/helper-functions/guards.ts
rafaelss95: Let's discuss it in #70.

src/rules/use-lifecycle-interface.ts
rafaelss95: Nice catch. I'll update it once #70 gets merged.

src/utils/helper-functions/utils.ts
timdeschryver: We have to thank you 🤗

src/rules/use-lifecycle-interface.ts
rafaelss95: Done!

rafaelss95:
Hey @timdeschryver, I've addressed the feedback (code part)... about the docs part, I was thinking in something like https://github.com/cartant/eslint-plugin-rxjs#rules, WDYT?

timdeschryver (approved):

timdeschryver:
@rafaelss95 yea, I was thinking the same thing :) 
I'll update the docs later, because I also might tweak the error level of the rules because most of them are set to `error` now which might be a bit too opionated :D

PR #68, opened 2021-01-31
test: sync select-style file name

rafaelss95:
While browsing the tests, I noticed that the file containing the tests for `select-style` was not synchronized with the name of the rule.


timdeschryver (approved):
👀 Thanks @rafaelss95

PR #70, opened 2021-02-03
refactor: enable strict mode

rafaelss95:


src/rules/avoid-dispatching-multiple-actions-sequentially.ts
```diff
@@ -34,41 +35,43 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
-          })
+            }),
+          )
         }
       },
     }
   },
 })
+
+function isStoreDispatchExpression(
```
rafaelss95, line 63: It was giving some compilation errors when I enabled strict mode, so I correct them and move it to a function in order to make it more readable.

src/utils/helper-functions/guards.ts
```diff
@@ -1,33 +1,39 @@
 import { AST_NODE_TYPES, TSESTree } from '@typescript-eslint/experimental-utils'
 
 export function isIdentifier(node: TSESTree.Node): node is TSESTree.Identifier {
-  return node && node.type === AST_NODE_TYPES.Identifier
+  return node.type === AST_NODE_TYPES.Identifier
```
rafaelss95, line 4: Hey Tim, about the nullity checks you mentioned [here](https://github.com/timdeschryver/eslint-plugin-ngrx/pull/67#discussion_r568498437)... it seems that they were unnecessary, since the received type does not include `undefined` or even `null`, but in any case let me know your opinion so I can update it accordingly.

timdeschryver: Oh weird, I can't remember if I added them here or if I copy-pasted it from another project where these could be `null`. 🤔 Let's keep it like this.

timdeschryver: After a closer look, this isn't required anymore because you checked the node before it reaches the guards.
e.g.

```diff
+ node.parent &&
+ isCallExpression(node.parent) &&
- isCallExpression(node.parent) &&
```

timdeschryver (approved):

PR #71, opened 2021-02-06
feat: add fixer for no-typed-store rule

rafaelss95:
This PR aims to add a ~fixer~ suggest (it can't be a fixer as it can produce invalid code as reported in https://github.com/ngrx/platform/issues/2780).

### DIFF
--- a/src/rules/no-typed-store.ts
+++ b/src/rules/no-typed-store.ts
@@ -5,7 +5,8 @@ import { typedStore, docsUrl } from '../utils'
 export const ruleName = 'no-typed-store'
 
 export const messageId = 'noTypedStore'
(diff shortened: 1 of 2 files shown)


timdeschryver (approved):
👍 Just one question, can't we use `fromFixture` for this ?
Just asking because I really like the squiglies to spot the rules.

rafaelss95:
> 👍 Just one question, can't we use `fromFixture` for this ?
> Just asking because I really like the squiglies to spot the rules.

AFAIK we can't test `suggestions` using `fromFixture` 😞

timdeschryver:
Eh, I thought I tried that before and it did work. But it seems like you're right... sorry 😅

PR #72, opened 2021-02-06
feat: add fixer for no-effect-decorator rule

rafaelss95:


src/utils/helper-functions/utils.ts
```diff
@@ -1,5 +1,6 @@
 import type { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils'
 import {
+  isCallExpression,
```
rafaelss95, line 3: @timdeschryver While doing this, I noticed that many of the guards we implemented here already exist in [**eslint-etc**](https://github.com/cartant/eslint-etc/blob/ee59baa6caf2ddcfd20cf398acabe8d75df4a41a/source/is.ts). If you agree, I can send a follow-up PR replacing these.

timdeschryver: I was waiting on the TS eslint package to create these guards.
They already have some at https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/experimental-utils/src/ast-utils/predicates.ts but not all of them :( 
Let's create a new issue for this and evaluate later what we want to do with it?

timdeschryver (approved):
Looks good to me @rafaelss95 , I added a commit to change the arrow syntax of an effect to a return statement. This is recommended because it's easier to spot TypeScript errors.
Let me know what you think about it, so we can merge this in, or revert the change.

src/utils/helper-functions/utils.ts
rafaelss95: Perfect. I'll raise an issue so we don't forget this.

rafaelss95:
> Looks good to me @rafaelss95 , I added a commit to change the arrow syntax of an effect to a return statement. This is recommended because it's easier to spot TypeScript errors.
> Let me know what you think about it, so we can merge this in, or revert the change.

I have to confess that I always prefer to omit `return`, but I'm okay with this also 👍

src/rules/no-effect-decorator.ts
```diff
@@ -18,20 +35,90 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
-      [effectDecorator](node: TSESTree.TSTypeReference) {
+      [classPropertyWithEffectDecorator](node: EffectDecorator) {
+        const isUsingEffectCreator =
+          isIdentifier(node.parent.value.callee) &&
+          node.parent.value.callee.name === createEffect
+
         context.report({
+          ...(isUsingEffectCreator
+            ? {
+                suggest: [
```
timdeschryver, line 58: One more question that I forgot to ask yesterday.
Could this also be a fix instead of a suggest?

rafaelss95: I went as a suggest because having `@Effect` and also `createEffect`, the fix (it is, the deletion of `@Effect`) could introduce a break in the expected behavior as, for some reason, they can have different configs setted, as for example:

```
@Effect({ dispatch: true })
readonly test$ = createEffect(() => {
  ...
}, { dispatch: false })
```

That being said, do you think that it would be interesting to add a comment explaining this?

timdeschryver: I think it can just be a fix. My reasoning is that it's probably just a leftover from a refactor, if both effects are still there it's probably already doing something unwanted 😅

PR #73, opened 2021-02-06
feat: add fixer for no-dispatch-in-effects

rafaelss95:
I went with a _suggest_ instead of _fix_ in this case because although it does not produce invalid code, it can cause some unexpected behavior if for example someone sets `dispatch: false` and still calls `store$.dispatch` manually (as in the test), there may be breaks in the expected flow.

### DIFF
--- a/src/rules/no-dispatch-in-effects.ts
+++ b/src/rules/no-dispatch-in-effects.ts
@@ -4,8 +4,11 @@ import { dispatchInEffects, docsUrl } from '../utils'
 
 export const ruleName = 'no-dispatch-in-effects'
 
-export const messageId = 'NoDispatchInEffects'
-export type MessageIds = typeof messageId
+export const noDispatchInEffects = 'noDispatchInEffects'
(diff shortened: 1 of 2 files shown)


timdeschryver (commented):
I don't think we can fix this automatically, the test only has a simple case but the `store.dispatch` could be nested, or it could even be called multiple times.

rafaelss95:
Yep, that's why I implemented it as a suggestion. I find suggestions really useful, because you hover to see the problem and check if the suggestion solves the problem properly at the same time. Let me know your thoughts on this.

timdeschryver:
I'm going to think on this a while longer.
While I see why we want to do it, I'm worried peoples will just use the suggestion and expect things to work.

timdeschryver:
I'm going to close this, since I'm still not convinced.
Thanks for the effort you've put int this @rafaelss95

rafaelss95:
No problem, Tim! To be honest, I no longer remembered that this PR was still open 😄

Issue #82, opened 2021-02-10
Refactor avoid-dispatching-multiple-actions-sequentially and use-selector-in-select


timdeschryver:
Coming from https://github.com/timdeschryver/eslint-plugin-ngrx/pull/80#issuecomment-776566485

PR #83, opened 2021-02-13
refactor(use-selector-in-select): implementation using esquery

timdeschryver:
Partially #82

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,8 +11,10 @@ module.exports = {
     ecmaVersion: 2018,
     sourceType: 'module',
     project: './tsconfig.eslint.json',
+    extraFileExtensions: ['.json'],
   },
   rules: {
     '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/no-unused-vars': ['error', { varsIgnorePattern: '_' }],
   },
 }
--- a/src/rules/select-style.ts
+++ b/src/rules/select-style.ts
@@ -3,7 +3,7 @@ import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
 import {
   docsUrl,
   pipeableSelect,
-  readNgRxStoreNameFromSettings,
+  readNgRxStoreName,
   storeSelect,
 } from '../utils'
 
@@ -50,7 +50,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   },
   defaultOptions: [{ mode: METHOD }],
   create: (context, [{ mode }]) => {
-    const storeName = readNgRxStoreNameFromSettings(context.settings)
+    const storeName = readNgRxStoreName(context)
 
     if (mode === METHOD) {
       return {
--- a/src/rules/select-style.ts
+++ b/src/rules/select-style.ts
@@ -1,8 +1,8 @@
 import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
 
 import {
+  createStoreSelectCallExpressionVisitors,
   docsUrl,
-  pipeableSelect,
   readNgRxStoreNameFromSettings,
   storeSelect,
 } from '../utils'
@@ -51,16 +51,19 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   defaultOptions: [{ mode: METHOD }],
   create: (context, [{ mode }]) => {
     return {
-      [pipeableSelect](node: TSESTree.CallExpression) {
-        if (mode === METHOD) {
-          context.report({
-            node,
-            messageId: methodSelectMessageId,
-          })
-        }
-      },
+      ...createStoreSelectCallExpressionVisitors(
+        readNgRxStoreNameFromSettings(context.settings),
+        (node) => {
+          if (mode === METHOD) {
+            context.report({
+              node,
+              messageId: methodSelectMessageId,
+            })
+          }
+        },
+      ),
       [storeSelect(readNgRxStoreNameFromSettings(context.settings))](
-        node: TSESTree.Identifier,
+        node: TSESTree.CallExpression,
       ) {
         if (mode === OPERATOR) {
           context.report({
--- a/src/rules/select-style.ts
+++ b/src/rules/select-style.ts
@@ -50,24 +50,25 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   },
   defaultOptions: [{ mode: METHOD }],
   create: (context, [{ mode }]) => {
-    return {
-      [pipeableSelect(readNgRxStoreNameFromSettings(context.settings))](node) {
-        if (mode === METHOD) {
+    const storeName = readNgRxStoreNameFromSettings(context.settings)
+
+    if (mode === METHOD) {
+      return {
+        [pipeableSelect(storeName)](node: TSESTree.CallExpression) {
           context.report({
             node,
             messageId: methodSelectMessageId,
           })
-        }
-      },
-      [storeSelect(readNgRxStoreNameFromSettings(context.settings))](
-        node: TSESTree.CallExpression,
-      ) {
-        if (mode === OPERATOR) {
-          context.report({
-            node,
-            messageId: operatorSelectMessageId,
-          })
-        }
+        },
+      }
+    }
+
+    return {
+      [storeSelect(storeName)](node: TSESTree.CallExpression) {
+        context.report({
+          node,
+          messageId: operatorSelectMessageId,
+        })
       },
     }
   },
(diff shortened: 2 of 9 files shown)


src/utils/selectors/index.ts
```diff
@@ -22,15 +24,32 @@ export const ngModuleProviders = `${ngModuleDecorator} ObjectExpression Property
...
+export const createStoreSelectCallExpressionVisitors = (
+  storeName: string,
+  visitor: (node: TSESTree.CallExpression) => void,
+) => {
+  return {
+    [pipeableSelectThis(storeName)]: visitor,
+    [pipeableSelect(storeName)]: visitor,
+  }
+}
+
```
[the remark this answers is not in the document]
timdeschryver: Neat! thanks

src/rules/select-style.ts
```diff
@@ -51,7 +51,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   defaultOptions: [{ mode: METHOD }],
   create: (context, [{ mode }]) => {
     return {
```
rafaelss95, line 53: To avoid reading store multiple times, we could read this only on initialization instead:

```suggestion
    const storeName = readNgRxStoreNameFromSettings(context.settings)

    return {
```

src/rules/select-style.ts
```diff
@@ -51,7 +51,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
-      [pipeableSelect](node: TSESTree.CallExpression) {
+      [pipeableSelect(readNgRxStoreNameFromSettings(context.settings))](node) {
```
rafaelss95, line 54: It looks like you removed the type annotation accidentally:

```suggestion
      [pipeableSelect(readNgRxStoreNameFromSettings(context.settings))](node: TSESTree.CallExpression) {
```

src/rules/use-selector-in-select.ts
```diff
@@ -34,68 +33,20 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
-            context.report({
-              node: n.node,
-              messageId,
-            })
-          }
-        })
-      },
+    return {
+      [pipeableSelect(readNgRxStoreNameFromSettings(context.settings))]: report,
+      [storeSelect(readNgRxStoreNameFromSettings(context.settings))]: report,
```
rafaelss95, line 49: The same I mentioned in `select-style`. Also, we could combine these selectors into one using interpolation. I was thinking of something like this (feel free to skip the usage of `filter` + `forEach` if you prefer `for-of` approach):

```suggestion
    const storeName = readNgRxStoreNameFromSettings(context.settings)

    return {
      [`${pipeableSelect(storeName)}, ${storeSelect(storeName)}`]({
        arguments: args,
      }: TSESTree.CallExpression) {
        args
          .filter((node) => isLiteral(node) || isFunctionExpressionLike(node))
          .forEach((node) =>
            context.report({
              node,
              messageId,
            }),
          )
      },
    }
```

src/utils/selectors/index.ts
```diff
@@ -1,3 +1,5 @@
+import { TSESTree } from '@typescript-eslint/experimental-utils'
+
 export const effectCreator = `ClassProperty[value.callee.name='createEffect']`
 
 export const effectDecorator = `Decorator[expression.callee.name='Effect']`
```
rafaelss95, line 5: it looks like the `import` isn't used anymore.

```suggestion
export const effectCreator = `ClassProperty[value.callee.name='createEffect']`
```

rafaelss95 (commented):
Some minor comments, but in general LGTM! `use-selector-in-select` is not only much cleaner now, but also more robust with less code ❤️

src/utils/selectors/index.ts
```diff
@@ -1,3 +1,5 @@
+import { TSESTree } from '@typescript-eslint/experimental-utils'
+
 export const effectCreator = `ClassProperty[value.callee.name='createEffect']`
```
rafaelss95, line 3: it looks like the `import` isn't used anymore.

```suggestion
export const effectCreator = `ClassProperty[value.callee.name='createEffect']`
```

src/rules/use-selector-in-select.ts
timdeschryver: Oh I keep forgetting this 🤦‍♂️

.eslintrc.js
```diff
@@ -11,8 +11,10 @@ module.exports = {
     ecmaVersion: 2018,
     sourceType: 'module',
     project: './tsconfig.eslint.json',
+    extraFileExtensions: ['.json'],
   },
   rules: {
     '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/no-unused-vars': ['error', { varsIgnorePattern: '_' }],
```
rafaelss95, line 18: Since we're configuring it as an `error`, shouldn't it be better to use [`noUnusedLocals`](https://www.typescriptlang.org/tsconfig#noUnusedLocals) and [`noUnusedParameters`](https://www.typescriptlang.org/tsconfig#noUnusedParameters) in `tsconfig.json` instead? I'm just asking because as I generally prefer to use TS flags on my personal projects, I'd love to see a second opinion on this :sweat_smile:

src/utils/helper-functions/typecheck.ts
```diff
@@ -1,10 +1,7 @@
 import { TSESTree } from '@typescript-eslint/experimental-utils'
 import { couldBeType } from 'tsutils-etc'
 
-function getParserServices(
-  // RuleContext<string, readonly unknown[]>>
-  context: Readonly<any>,
-) {
+function getParserServices(context: Readonly<any>) {
```
rafaelss95, line 4: I guess it's violating the `@typescript-eslint/no-explicit-any` rule, no? You could use this instead:

```suggestion
function getParserServices(
  context: TSESLint.RuleContext<string, readonly unknown[]>,
) {
```

src/utils/helper-functions/typecheck.ts
```diff
@@ -17,10 +14,7 @@ function getParserServices(
   return context.parserServices
 }
 
-export function typecheck(
-  // RuleContext<string, readonly unknown[]>>
-  context: Readonly<any>,
-) {
+export function typecheck(context: Readonly<any>) {
```
rafaelss95, line 17:
```suggestion
function typecheck(
  context: TSESLint.RuleContext<string, readonly unknown[]>,
) {
```

src/rules/select-style.ts
```diff
@@ -50,24 +50,25 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   },
   defaultOptions: [{ mode: METHOD }],
   create: (context, [{ mode }]) => {
-    return {
-      [pipeableSelect](node: TSESTree.CallExpression) {
-        if (mode === METHOD) {
+    const storeName = readNgRxStoreNameFromSettings(context.settings)
+
+    if (mode === METHOD) {
```
rafaelss95, line 55: Nice 🚀

src/utils/helper-functions/typecheck.ts
```diff
@@ -1,10 +1,7 @@
 import { TSESTree } from '@typescript-eslint/experimental-utils'
```
rafaelss95, line 1:
```suggestion
import type { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils'
```

.eslintrc.js
timdeschryver: I like the eslint option because we can ignore `_` variables.
I'm not sure if this is also the case with the TS flags.

src/utils/helper-functions/typecheck.ts
timdeschryver: I was looking for these types, but couldn't find them, only in the dist folder.
Thanks!
I'll also use this to grab the store name from the settings, instead of passing the settings down to `readNgRxStoreNameFromSettings`

src/utils/helper-functions/typecheck.ts
timdeschryver: TIL, thanks again :)

.eslintrc.js
rafaelss95: Hmm yes, AFAIK we can't ignore underscored **variables** with them, only parameters.

src/utils/helper-functions/typecheck.ts
rafaelss95: Happened to me also with `RuleFixer` and I even opened an [issue](https://github.com/typescript-eslint/typescript-eslint/issues/2931) there. Since that discovery, I've been using it :)

rafaelss95 (approved):
LGTM! 🥳

PR #85, opened 2021-02-18
refactor: smart store name detection

timdeschryver:
@rafaelss95 what do you think of this? I want to hear your opinion first before fixing all the test, if we agree this is a good approach.
This PR, detects the name given to the store variable (instead of using `store`, `store$` or a custom defined name in the settings).
It does this by looking for the store import and then by looking for the declaration,  this all before any esquery is executed.


rafaelss95:
Hey @timdeschryver, sorry for the late comment 😞

About this approach, I like it as it seems to give much more freedom than the current one, but before proceed, if possible, 
I'd like to ask you something: the fact that we get the `sourceCode` and all the `declaredVariables` for these rules could somehow cause some negative performance impact or would it be insignificant?

timdeschryver:
I'm certain that this will be slower, but that the impact will be negligible.
By looking at some of the eslint rules, I noticed [these rules](https://github.com/eslint/eslint/search?q=getsourcecode) frequently invoke `getSourceCode`, it also [seems like it's cached](https://github.com/eslint/eslint/blob/master/lib/linter/linter.js#L863).

Don't worry about "late replies", I'm happy that you make some time to contribute to this project 🙌

timdeschryver:
Update: while updating all the tests I noticed the esqueries were throwing errors (invalid queries) when there was not store name provided. So, for these rules we can "disable" them when there's no store injected. With this change, these rules might be faster than before 🙂

src/utils/helper-functions/utils.ts
```diff
@@ -83,3 +86,48 @@ export function getDecoratorArgument({ expression }: TSESTree.Decorator) {
...
+          return ref.identifier.parent.parent.parent.name
+        }
+        return ''
+      })
+      .find(Boolean)
+    storeName = storeImportName || ''
+  }
+
+  return storeName
+}
```
rafaelss95, line 133: I was thinking we could just use the `findImportDeclarationNode` function present in this file... also, we could change the signature to be  `string | undefined`, instead of returning an empty string.
What do you think?

```suggestion
export function findNgRxStoreName(
  context: TSESLint.RuleContext<string, readonly unknown[]>,
): string | undefined {
  const { ast } = context.getSourceCode()
  const storeImportSpecifier = findImportDeclarationNode(ast, '@ngrx/store')

  if (!storeImportSpecifier) return undefined

  const [{ references }] = context.getDeclaredVariables(storeImportSpecifier)

  return references
    .map(({ identifier: { parent } }) => {
      if (
        parent &&
        isTSTypeReference(parent) &&
        parent.parent &&
        isTSTypeAnnotation(parent.parent) &&
        parent.parent.parent &&
        isIdentifier(parent.parent.parent)
      ) {
        return parent.parent.parent.name
      }
      return undefined
    })
    .find(Boolean)
}
```

timdeschryver: 🚀

PR #88, opened 2021-02-20
fix(ngrx-avoid-dispatching-multiple-actions-sequentially): only check violations in their own scope

### LINKED ISSUE #86: False-positive for ngrx-avoid-dispatching-multiple-actions-sequentially in subscriptions
Having onNgInit of my component like this:

```
ngOnInit(): void {
    this.route.params.subscribe(({ id }) => {
      this.id = id;
      if (id) {
        this.store.dispatch(
          load({
            id: this.id,
            page: this.page,
            per_page: this.limit,
            ransack: this.conditions
          })
        );
      }
    });

    this.form
      .get('search')
      .valueChanges.pipe(
        debounceTime(1000),
        takeWhile(() => this.alive)
      )
      .subscribe(value => {
        if (value.length > 0) {
          this.conditions = {
            ...this.conditions,
            code_or_name_cont: value
          };
        } else {
          const { code_or_name_cont, ...conditions } = this.conditions;
          this.conditions = conditions;
        }
        this.store.dispatch(
          load({
            id: this.id,
            page: this.page,
            per_page: this.limit,
            ransack: this.conditions
          })
        );
      });
    this.store.pipe(select(getActiveId), take(1)).subscribe(id => {
      if (Number(this.id) !== id) {
        this.store.dispatch(changeActive({ id: id }));
      }
    });
  }
```
leads to false-positive ngrx-avoid-dispatching-multiple-actions-sequentially on second and third dispatch.

timdeschryver:
Closes #86

### DIFF
--- a/src/rules/avoid-dispatching-multiple-actions-sequentially.ts
+++ b/src/rules/avoid-dispatching-multiple-actions-sequentially.ts
@@ -9,7 +9,7 @@ import {
 
 export const ruleName = 'avoid-dispatching-multiple-actions-sequentially'
 
-export const messageId = 'AvoidDispatchingMultipleActionsSequentially'
+export const messageId = 'avoidDispatchingMultipleActionsSequentially'
 export type MessageIds = typeof messageId
 
 type Options = []
@@ -36,7 +36,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
       TSESTree.CallExpression[]
     >()
     return {
-      [`${actionDispatch(readNgRxStoreName(context))}`](
+      [actionDispatch(readNgRxStoreName(context))](
         node: TSESTree.CallExpression,
       ) {
         if (
@@ -52,7 +52,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
           ])
         }
       },
-      ['Program:exit']() {
+      'Program:exit'() {
         dispatchExpressionsWithSameParent.forEach((dispatches) => {
           if (dispatches.length > 1)
             dispatches.forEach((node) =>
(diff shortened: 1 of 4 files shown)


src/rules/avoid-dispatching-multiple-actions-sequentially.ts
```diff
@@ -26,13 +31,36 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+        ) {
+          const nodes =
+            dispatchExpressionsWithSameParent.get(node.parent.parent) || []
+          dispatchExpressionsWithSameParent.set(node.parent.parent, [
+            ...nodes,
+            node,
+          ])
+        }
+      },
+      ['Program:exit']() {
```
rafaelss95, line 55:
```suggestion
      'Program:exit'() {
```

src/rules/avoid-dispatching-multiple-actions-sequentially.ts
```diff
@@ -26,13 +31,36 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+        dispatchExpressionsWithSameParent.forEach((dispatches) => {
+          if (dispatches.length > 1)
+            dispatches.forEach((node) =>
+              context.report({
+                node,
+                messageId,
```
rafaelss95, line 61: While reading this file I just noticed a small typo in the `messageId`:

https://github.com/timdeschryver/eslint-plugin-ngrx/blob/4d22e142cc4465f75376d5a2fc5e4a991b7e11c9/src/rules/avoid-dispatching-multiple-actions-sequentially.ts#L7

... it starts with a capitalized letter and all other `messageId` starts with a lowercase letter. Would it be good to change it to maintain the standard? :sweat_smile:

rafaelss95 (approved):
Just a few suggestions, but LGTM.

<sup>Sadly that the sibling selector isn't enough to match all cases 😞 </sup>

PR #95, opened 2021-03-22
docs(avoid-dispatching-multiple-actions-sequentially): Create docs

Armenvardanyan95:


docs/rules/avoid-dispatching-multiple-actions-sequentially.md
```diff
@@ -1 +1,60 @@
...
+
+  constructor(
+    private readonly actions$: Actions,
+  ) {}
+}
+```
+
+## Further reading
+
+- [Good Action Hygiene with NgRx Mike Ryan](https://www.youtube.com/watch?v=JmnsEvoy-gY)
```
timdeschryver, line 60: This can be left out.

```suggestion
```

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
```diff
@@ -1 +1,60 @@
 # avoid-dispatching-multiple-actions-sequentially
+
+> This rule promotes disallows dispatching multiple actions one after another.
```
timdeschryver, line 3: We try to keep this line equal to the [rule's description](https://github.com/timdeschryver/eslint-plugin-ngrx/blob/master/src/rules/avoid-dispatching-multiple-actions-sequentially.ts#L23).

I like your description a lot more, so could you also update the [description of the rule]((https://github.com/timdeschryver/eslint-plugin-ngrx/blob/master/src/rules/avoid-dispatching-multiple-actions-sequentially.ts#L23)) please?
 
```suggestion
> This rule disallows dispatching multiple actions one after another.
```

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
```diff
@@ -1 +1,60 @@
...
+
+An action should be an event which abstracts away the details of store internals.
+An action can be a composition of several events, in which case the developer might be tempted to dispatch several actions in sequence. But a better approach is t dispatch one "combining" action, which exactly describes what that event entails, and then map it to several other actions in an Effect
+
+Examples of **incorrect** code for this rule:
+
+```ts
+export class Component implement OnInit {
+  constructor(
+    private readonly store: Store<AppState>,
```
timdeschryver, line 15:
```suggestion
    private readonly store: Store,
```

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
```diff
@@ -1 +1,60 @@
...
+}
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+// in component code:
+export class Component implement OnInit {
+  constructor(
+    private readonly store: Store<AppState>,
```
timdeschryver, line 32:
```suggestion
    private readonly store: Store,
```

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
```diff
@@ -1 +1,60 @@
...
+```ts
+export class Component implement OnInit {
+  constructor(
+    private readonly store: Store<AppState>,
+  ) {}
+
+  ngOnInit() {
+    this.store.dispatch(loadEmployeeList());
+    this.store.dispatch(loadCompanyList());
+    this.store.dispatch(cleanData());
```
timdeschryver, line 21: Should we highlight that this is the problem here?

```suggestion
   // ⚠ multiple actions dispatched
    this.store.dispatch(loadEmployeeList());
    this.store.dispatch(loadCompanyList());
    this.store.dispatch(cleanData());
```

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
```diff
@@ -1 +1,60 @@
...
+export class Effects {
+
+  loadInitialDataForComponent$ = createEffect(() => this.actions.pipe(
+    ofType(componentLoaded),
+    switchMap(() => [
+      loadEmployeeList();
+      loadCompanyList();
+      cleanData();
+    ]),
+  ));
```
timdeschryver, line 50: Actually, doing this is also a bad-practice for the same reasons as this rule.
An Effect should also result in one action that's dispatched, instead of 3.

We recommend to listen to the same action in 3 effects. For example a `loadEmployeeList` effect that makes the http request and dispatches a succes, or a failed action back to the store

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
```diff
@@ -1 +1,60 @@
 # avoid-dispatching-multiple-actions-sequentially
+
+> This rule promotes disallows dispatching multiple actions one after another.
+
+## Rule Details
+
+An action should be an event which abstracts away the details of store internals.
+An action can be a composition of several events, in which case the developer might be tempted to dispatch several actions in sequence. But a better approach is t dispatch one "combining" action, which exactly describes what that event entails, and then map it to several other actions in an Effect
```
timdeschryver, line 8:
```suggestion
An action can be a composition of several events, in which case the developer might be tempted to dispatch several actions in sequence. But a better approach is t dispatch one "combining" action, which exactly describes what that event entails.
```

timdeschryver (commented):
Thanks for adding these docs Armen!
I left some comments, feel free to ping me if needed.

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
Armenvardanyan95: Wait, really?  Never realized this is a bad practice. How come? Is it possibly because some parts of the state might not be loaded?

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
Armenvardanyan95: Yes, I agree

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
Armenvardanyan95: I see, I will change the example

Armenvardanyan95: Done ✔️

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
```diff
@@ -1 +1,66 @@
 # avoid-dispatching-multiple-actions-sequentially
+
+> This rule disallows dispatching multiple actions one after another.
+
+## Rule Details
+
+An action should be an event which abstracts away the details of store internals.
+An action can be a composition of several events, in which case the developer might be tempted to dispatch several actions in sequence. But a better approach is t dispatch one "combining" action, which exactly describes what that event entails.
```
timdeschryver, line 8:
```suggestion
An action can be a composition of several events, in which case the developer might be tempted to dispatch several actions in sequence. But a better approach is to dispatch one "combining" action, which exactly describes what that event entails.
```

timdeschryver (approved):

docs/rules/avoid-dispatching-multiple-actions-sequentially.md
timdeschryver: Yea, I'll create the docs for it and send you the link 😉

PR #108, opened 2021-04-15
docs: no multiple actions in effects

Armenvardanyan95:
Added a rule doc

### DIFF
--- a/docs/rules/no-multiple-actions-in-effects.md
+++ b/docs/rules/no-multiple-actions-in-effects.md
@@ -5,7 +5,7 @@
 ## Rule Details
 
 An Effect should map one event (action) to a single other action.
-An action can result in several other Effects being triggered, or multiple changes in the reducer, in which case the developer might be tempted to map an Effect to several actions. A more understandable approach is to dispatch one "combining" action that describes what happened, rather than multiple actions.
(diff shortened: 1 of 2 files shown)


docs/rules/no-multiple-actions-in-effects.md
```diff
@@ -1 +1,76 @@
 # no-multiple-actions-in-effects
+
+> This rule disallows mapping to multiple actions in a single Effect.
+
+## Rule Details
+
+An Effect should map one event (action) to a single other action.
+An action can result in several other Effects being triggered, or multiple changes in the reducer, in which case the developer might be tempted to map an Effect to several actions. A more understandable approach is to dispatch one "combining" action that describes what happened, rather than multiple actions.
```
timdeschryver, line 8:
```suggestion
An action can result in several other Effects being triggered, or multiple changes in the reducer, in which case the developer might be tempted to map an Effect to several actions. A more understandable approach is to dispatch one "combining" action that describes what happened (an event), rather than multiple actions.
```

Armenvardanyan95: Agreed, this better describes what exactly an action is

timdeschryver (approved):

PR #109, opened 2021-04-18
feat: add prefer-effect-callback-in-block-statement

timdeschryver:


docs/rules/prefer-effect-callback-in-block-statement.md
```diff
@@ -0,0 +1,38 @@
...
+## Rule Details
+
+This rule prefers that the callback of an effect is a block statements.
+This makes it easier to troubleshoot type errors, for when example an RxJS operator isn't imported.
+
+Examples of **incorrect** code for this rule:
+
+```ts
+class Effect {
+  effectNOK = createEffect(() =>
```
Armenvardanyan95, line 14: Maybe change to `effectNotOk`? I was wondering for a minute if maybe NOK is some NgRx terminology  unknown to me 😄

timdeschryver: I'm going to merge this in, I believe other are also using NOK.
I'll refactor that later and all in one PR 🙂
Thanks for the review!

PR #111, opened 2021-04-19
docs(no-typed-store): add docs

Armenvardanyan95:
Created a doc for the rule


docs/rules/no-dispatch-in-effects.md
```diff
@@ -1 +1,60 @@
 # no-dispatch-in-effects
+
+> This rule disallows dispatching actions as side effects in NgRx Effects.
+
+## Rule Details
+
+An effect should handle actions and map it to a singular action. Each effect should be clear and concise and must be readily understandable as to what it does affect. Dispatching from inside an effect instead (or together with) mapping it to a new a action can result in painfully hard-to-find bugs and is generally considered a bad practice.
```
jsaguet-betclic, line 7:
```suggestion
An effect should handle actions and map them to a singular action. Each effect should be clear and concise and must be readily understandable as to what it does affect. Dispatching from inside an effect instead of (or together with) mapping it to a new a action can result in painfully hard-to-find bugs and is generally considered a bad practice.
```

docs/rules/no-dispatch-in-effects.md
```diff
@@ -1 +1,60 @@
...
+
+```ts
+export class Effects {
+
+  loadData$ = createEffect(() => this.actions.pipe(
+    ofType(loadData),
+    exhaustMap(() => this.dataService.getData().pipe(
+      tap(response => {
+        // ⚠ dispatching another action from an effect
+        if (response.condition) {
```
jsaguet-betclic, line 19: isn't this condition unnecessary ? The example should be kept as minimal as possible imho

docs/rules/no-dispatch-in-effects.md
jsaguet, line 7:
```suggestion
An effect should handle actions and map them to a singular action. Each effect should be clear and concise and must be readily understandable as to what it does affect. Dispatching from inside an effect instead of (or together with) mapping it to a new a action can result in painfully hard-to-find bugs and is generally considered a bad practice.
```

docs/rules/no-typed-store.md
```diff
@@ -1 +1,36 @@
 # no-typed-store
+
+> This rule disallows providing typings to the generic Store type.
+
+## Rule Details
+
+Typing the global `Store` is redundant, and selectors are type safe, so mentioning the state interface while injecting the store is unnecessary.
```
timdeschryver, line 7:
```suggestion
Typing the global `Store` is redundant because selectors are type safe, so adding the generic state interface while injecting the store is unnecessary.
```

timdeschryver (approved):
Thanks again @Armenvardanyan95 , and thank you @jsaguet for the review!

PR #112, opened 2021-04-19
docs/no duplicate action in reducer

Armenvardanyan95:
Created a rule doc page


timdeschryver (commented):
Oops, I think that you don't have the lastest version of `master`.
This rule got removed a while ago because this is now valid syntax in NgRx.
I removed the docs in https://github.com/timdeschryver/eslint-plugin-ngrx/commit/d01be5f51cffb0b63370062831359117a64e0488#diff-7d3bdafbe1a7ebe50790ece5154a9673295b02f6f60b3ed8dbdde59ebeb0778f

PR #116, opened 2021-04-26
docs(no-effect-decorator-and-creator): add docs

Armenvardanyan95:
Created the rule doc


docs/rules/no-effect-decorator-and-creator.md
```diff
@@ -1 +1,39 @@
 # no-effect-decorator-and-creator
+
+> This rule disallows using the `@Effect` decorator and the `createEffect` function simultaneously
+
+## Rule Details
+
+There are two ways we can register an Effect in NgRx. One is using the `@Effect` decorator, the other is with the help of the `createEffect` function. Using both simultaneously will result in the Effect being registered twice, and the side-effect will be performed twice every time the corresponding action is dispatched.
```
timdeschryver, line 7:
```suggestion
There are two ways we can register an Effect in NgRx. One is using the `@Effect` decorator (this is currently deprecated), the other is with the help of the `createEffect` function. Using both simultaneously will result in the Effect being registered twice, and the side-effect will be performed twice every time the corresponding action is dispatched.
```

docs/rules/no-effect-decorator-and-creator.md
```diff
@@ -1 +1,39 @@
...
+
+Examples of **correct** code for this rule:
+
+```ts
+export class Effects {
+
+  loadData$ = createEffect(() => this.actions.pipe(
+    ofType(loadData),
+    // performing the side effect
+  ));
```
timdeschryver, line 33:
```suggestion
  loadData$ = createEffect(() => {
    return this.actions.pipe(
      ofType(loadData),
      // performing the side effect
    ))
  };
```

docs/rules/no-effect-decorator-and-creator.md
Armenvardanyan95: Very nice! I thought about adding a hint that using `@Effect` is not good either, but somehow forgot

docs/rules/no-effect-decorator-and-creator.md
Armenvardanyan95: I keep writing "good" examples that violate other rules 😁

docs/rules/no-effect-decorator-and-creator.md
```diff
@@ -1 +1,41 @@
...
+## Rule Details
+
+There are two ways we can register an Effect in NgRx. One is using the `@Effect` decorator (this is currently deprecated), the other is with the help of the `createEffect` function. Using both simultaneously will result in the Effect being registered twice, and the side-effect will be performed twice every time the corresponding action is dispatched.
+
+Examples of **incorrect** code for this rule:
+
+```ts
+export class Effects {
+
+  @Effect() loadData$ = createEffect(() => this.actions.pipe(
```
timdeschryver, line 14:
```suggestion
  @Effect() loadData$ = createEffect(() => this.actions$.pipe(
```

docs/rules/no-effect-decorator-and-creator.md
```diff
@@ -1 +1,41 @@
...
+}
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+export class Effects {
+
+  loadData$ = createEffect(() => {
+    return this.actions.pipe(
```
timdeschryver, line 31:
```suggestion
    return this.actions$.pipe(
```

timdeschryver (approved):

PR #118, opened 2021-04-27
docs(select-style): add docs

Armenvardanyan95:
Add a rule doc


docs/rules/select-style.md
```diff
@@ -1 +1,25 @@
 # select-style
+
+> This rule enforces consistent usage of `select` method or `select` operator on the store.
+
+## Rule Details
+
+There are two ways of selecting data from the store, either by using the `this.store.select(selectorFn)` method, or by using the `this.store.pipe(select(selectorFn))` operator. Either way is considered correct (although the first way is preferred as it requires less code and no importing of the operator), but it is important to keep things consistent, so this rule will disallow using both across the same codebase.
```
timdeschryver, line 7:
```suggestion
There are two ways of selecting data from the store, either by using the `this.store.select(selectorFn)` method, or by using the `this.store.pipe(select(selectorFn))` operator. Either way is considered correct , although the first way is preferred as it requires less code and it doesn't require the need to import the `selector` operator. 

Because it's important to keep things consistent, this rule disallows using both across the same codebase.
```

timdeschryver (commented):
I think it's als important to add a note that it's configurable and how to change it.

Something familiar to 
https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-explicit-assert.md#options

Armenvardanyan95:
@timdeschryver thanks, will add it next

docs/rules/select-style.md
Armenvardanyan95: Agreed, thanks

timdeschryver (approved):

timdeschryver:
Thank for another PR @Armenvardanyan95 !

PR #122, opened 2021-05-03
feat(prefer-concat-latest-from): add new rule prefer-concat-latest-from

### LINKED ISSUE #100: 💡 New Rule: prefer concatLatestFrom
With version 11 NgRx has `concatLatestFrom` lazy operator which should be used instead of eager operators `withLatestFrom`. This would be applicable for effects and if possible for components/services.

Documentation on `concatLatestFrom`: https://ngrx.io/guide/effects/operators#concatlatestfrom

timdeschryver:
Closes #100


docs/rules/prefer-concat-latest-from.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+class Effect {
+  detail$ = createEffect(() => {
+    return this.actions.pipe(
+      ofType(ProductDetailPage.loaded),
+      concatLatestFrom(this.store.select(selectProducts)),
```
minijus, line 32: It should be
`concatLatestFrom(() => this.store.select(selectProducts)),`

PR #124, opened 2021-05-03
docs(prefer-inline-action-props): Add rule doc

Armenvardanyan95:
Added a rule document


docs/rules/prefer-inline-action-props.md
```diff
@@ -1 +1,23 @@
 # prefer-inline-action-props
+
+> This rule enforces using inline props when creating actions with the `createAction` function instead of named interfaces/types.
+
+## Rule Details
+
+Lots of actions in an NgRx codebase have props (the action `payload`), and we need to define its type when creating an action. It might seem better to use named interfaces or types when defining those types, but in reality it will obscure their meaning to the developer using them. Actions props are essentially like function arguments, and the function caller needs to know exactly what type of data to provide (which will result in better IDE experience).
```
timdeschryver, line 7:
```suggestion
Lots of actions in an NgRx codebase have props, and we need to define the props type of an action when the action is defined. It might seem better to use named interfaces or types while defining those types, but in reality, it will obscure their meaning to the developer using them. Actions props are essentially like function arguments, and the function caller needs to know exactly what type of data to provide (which results in better IDE experience).

Note: some property names are not allowed to be used, such as `type`
```

timdeschryver (approved):

PR #125, opened 2021-05-04
docs(on-function-explicit-return-type): add rule doc

Armenvardanyan95:
Added a rule doc page


docs/rules/on-function-explicit-return-type.md
```diff
@@ -1 +1,43 @@
 # on-function-explicit-return-type
+
+> This rule disallows using the `on` reducer function without providing an explicit return type.
+
+## Rule Details
+
+When we use the `on` function to create reducers, we usually copy the state into a new object, and then add the properties that are being modified after that certain action. This may result in unexpected typing problems: we can add new properties into the state that did not exist previously, and TypeScript would not infer the problem. The solution is to provide an explicit return type to the `on` function callback.
+
+This rule is opt-in.
```
timdeschryver, line 9: I don't think we should add those notes here, I've recently created more pre-configured configs and I will document this in the readme.

```suggestion
```

docs/rules/on-function-explicit-return-type.md
```diff
@@ -1 +1,43 @@
 # on-function-explicit-return-type
+
+> This rule disallows using the `on` reducer function without providing an explicit return type.
+
+## Rule Details
+
+When we use the `on` function to create reducers, we usually copy the state into a new object, and then add the properties that are being modified after that certain action. This may result in unexpected typing problems: we can add new properties into the state that did not exist previously, and TypeScript would not infer the problem. The solution is to provide an explicit return type to the `on` function callback.
```
timdeschryver, line 7:
```suggestion
When we use the `on` function to create reducers, we usually copy the state into a new object, and then add the properties that are being modified after that certain action. This may result in unexpected typing problems, we can add new properties into the state that did not exist previously. TypeScript doesn't see this as a problem, and might change the state's interface. The solution is to provide an explicit return type to the `on` function callback.
```

docs/rules/on-function-explicit-return-type.md
Armenvardanyan95: Sure, will take note for any future rule docs

timdeschryver (approved):

PR #129, opened 2021-05-07
Update argument passed to concatLatestFrom in documentation

david-shortman:
- Corrects the argument to be an anonymous function


timdeschryver:
@all-contributors please add @david-shortman  for docs

PR #145, opened 2021-05-26
docs(no-multiple-actions-in-effects): satisfy other rules

dangrussell:
examples satisfy prefer-effect-callback-in-block-statement

Also removed "$" from constructor parameters

### DIFF
--- a/docs/rules/no-multiple-actions-in-effects.md
+++ b/docs/rules/no-multiple-actions-in-effects.md
@@ -12,7 +12,7 @@ Examples of **incorrect** code for this rule:
 ```ts
 export class Effects {
   loadEmployeeList$ = createEffect(() => {
-    return this.actions.pipe(
+    return this.actions$.pipe(
       ofType(componentLoaded),
       exhaustMap(() =>
         this.dataService.loadEmployeeList().pipe(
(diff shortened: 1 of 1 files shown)


timdeschryver (commented):
Hi, thanks to align the rules in this PR!
The examples in NgRx are all using `actions$`, so I would like to keep this aligned (I think it could otherwise even be a warning if you're using the rxjs eslint rules).
Do you want to update the docs so they use `actions$`?

dangrussell:
> Do you want to update the docs so they use `actions$`?

Will do!

timdeschryver:
@dangrussell I'm sorry to be a pain, but this would als mean that we should update the usage of `this.actions` in the examples. These should also be renamed to `this.actions$`.

dangrussell:
> @dangrussell I'm sorry to be a pain, but this would als mean that we should update the usage of `this.actions` in the examples. These should also be renamed to `this.actions$`.

Ah, I see. I was restoring the previous names. I will make them all use the $. One more commit coming :)

timdeschryver (approved):

timdeschryver:
@all-contributors please add @dangrussell  for docs

PR #162, opened 2021-07-17
fix(avoid-combining-selectors): detect violation with select pipe

### LINKED ISSUE #75: [avoid-combining-selectors] Unexpected behaviors
Similar to #74,

1. It doesn't report when we use something other than `store`:

`readonly something$ = combineLatest(this.store$.select(AuthSelectors.selectUser), this.store$.select(AuthSelectors.something))`

2. It doesn't report when we use `store.pipe` instead of `store.select`:

`readonly something$ = combineLatest(this.store$.pipe(select(AuthSelectors.selectUser)), this.store$.select(AuthSelectors.something))`

timdeschryver:
Closes #75

### DIFF
--- a/src/rules/store/avoid-combining-selectors.ts
+++ b/src/rules/store/avoid-combining-selectors.ts
@@ -40,14 +40,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
         node: TSESTree.CallExpression,
       ) {
         const [, ...violations] = node.arguments.filter(
-          (p) =>
-            isCallExpression(p) &&
-            isMemberExpression(p.callee) &&
-            isMemberExpression(p.callee.object) &&
-            isIdentifier(p.callee.object.property) &&
-            p.callee.object.property.name === storeName &&
-            isIdentifier(p.callee.property) &&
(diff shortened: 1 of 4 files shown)


src/rules/store/avoid-combining-selectors.ts
```diff
@@ -40,14 +40,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
         node: TSESTree.CallExpression,
```
rafaelss95, line 40: I think you could achieve the same result using the existent selectors, no?

```suggestion
    const pipeableOrStoreSelect = `:matches(${storeExpression(
      storeName,
    )}[callee.property.name='pipe']:has(CallExpression[callee.name='select']), ${storeSelect(
      storeName,
    )})`

    return {
      [`CallExpression[callee.name='combineLatest'][arguments.length>1] ${pipeableOrStoreSelect} ~ ${pipeableOrStoreSelect}`](
        node: TSESTree.CallExpression,
      ) {
        context.report({
          node,
          messageId,
        })
      },
```

timdeschryver: Nice! 🤩

PR #169, opened 2021-08-17
feat(prefix-selectors-with-select): add new rule

### LINKED ISSUE #168: New Rule: Use `select` In Selector names

https://redux.js.org/style-guide/style-guide#name-selector-functions-as-selectthing
Which is inline with NgRx ex https://github.com/ngrx/platform/pull/2151

```ts
export const users = createFeatureSelector<AppState, UsersState>('users');
```

## A snippet of correct code

```ts
export const selectUsers = createFeatureSelector<AppState, UsersState>('users');
```

- [x] Yes (Assistance is provided if you need help submitting a pull request)
- [ ] No

prince-chrismc:
Closes #168


src/utils/selectors/index.ts
```diff
@@ -55,3 +55,8 @@ export const createEffectBody = `${createEffectExpression} > ArrowFunctionExpres
 export const effectsImplicitReturn = `${mapOperatorsExpression} > ArrowFunctionExpression > ArrayExpression, ${mapToOperatorsExpression} ArrayExpression`
 
 export const effectsReturn = `${mapOperatorsExpression} ReturnStatement`
+
+const selectorsNames = 'createSelector|createFeatureSelector' // How to handle createSelectorFactory???
+export const selectorsExpression = `CallExpression[callee.name=/${selectorsNames}/]`
+const memorizedSelectorType = 'MemoizedSelector|MemoizedSelectorWithProps'
+export const memorizedSelector = `VariableDeclarator[id.typeAnnotation.typeAnnotation.typeName.name=/^${memorizedSelectorType}$/] > ObjectExpression > Property`
```
prince-chrismc, line 62: This is my first attempt at eslint Typescript rules so I am not sure how to get a variable based on type

Any help would be appreciated

timdeschryver: Hi, first of all thank you for your contribution! 
I'll take a look later this week for this (I'm using https://astexplorer.net/ for most of this, but it seems like it's currently a bit broken).

I think to make things simpler; we should only scan for usages of `createSelector` (and not for variables that have `: MemoizedSelector<any, any>` types).

I'm also not sure if we can write this rule with a single query, we could scan for `createSelector` instead and then go up in the tree to find the variable declarator.

prince-chrismc: Oh that's an interesting tool! I would not be surprised it needs to be more complicated.

If I have any luck I'll report back as well :)

prince-chrismc:
I had some luck doing it a little differently ... let me know what you think =)

src/rules/store/select-in-selector-names.ts
```diff
@@ -0,0 +1,76 @@
...
+    function hasKeyword(name: string): boolean {
+      // console.log(name.toLowerCase().startsWith(includedKeyword))
+      return name.toLowerCase().startsWith(includedKeyword)
+    }
+
+    return {
+      [`VariableDeclaration[declarations.length>0]`](
+        node: TSESTree.VariableDeclaration,
+      ) {
+        for (const dec of node.declarations) {
```
timdeschryver, line 43: I think we can query for `VariableDeclarator` (instead of `VariableDeclaration[declarations.length>0]`).
This way, we don't have to iterate over de declarators.

src/rules/store/select-in-selector-names.ts
```diff
@@ -0,0 +1,76 @@
...
+                  .typeName as TSESTree.Identifier).name as string,
+              ) === null
+            ) {
+              if (!hasKeyword(id.name)) {
+                context.report({ node: id, messageId })
+              }
+            }
+          } else {
+            // console.log(dec.init)
+            if (dec.init?.type === AST_NODE_TYPES.CallExpression) {
```
timdeschryver, line 60: This is correct, but I'd prefer to use our [custom utilies](https://github.com/timdeschryver/eslint-plugin-ngrx/blob/main/src/utils/helper-functions/guards.ts#L11).
The benefit of using them is that they are typed.

src/rules/store/select-in-selector-names.ts
```diff
@@ -0,0 +1,76 @@
...
+              if (
+                /^${selectorsNames}$/.exec(
+                  (dec.init.callee as TSESTree.Identifier).name,
```
timdeschryver, line 63: It would be better to be a bit safer here.
Instead of using a type assert, I would prefer to also use the custom utilities.

e.g.

```ts
//                dec.init.callee knows it's an identifier now 👇
if(isIdentifier(dec.init.callee) && /^${selectorsNames}$/.exec(dec.init.callee)){
...
}
```

src/rules/store/select-in-selector-names.ts
```diff
@@ -0,0 +1,76 @@
...
+    return {
+      [`VariableDeclaration[declarations.length>0]`](
+        node: TSESTree.VariableDeclaration,
+      ) {
+        for (const dec of node.declarations) {
+          const id: TSESTree.Identifier = (dec.id as any) as TSESTree.Identifier
+          // console.log(id.name) //, dec.id.typeAnnotation?.typeAnnotation)
+          if (dec.id.typeAnnotation?.typeAnnotation) {
+            if (
+              /^${memorizedSelectorTypes}$/.exec(
```
timdeschryver, line 48: Instead of `exec`, it's better to use `test` this will return a `true` or a `false` and not the match.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

timdeschryver (commented):
This seems good to me 👍
I did leave a few comments about things that can be improved. 
If you convert the draft to a "real" PR, don't forget to cleanup the console.log statements in comments, and I think there are also a few variables that aren't used anymore.

src/rules/store/select-in-selector-names.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type Options = []
+
+export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
+  name: path.parse(__filename).name,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      category: 'Best Practices',
+      description: 'Name Selector Functions as selectThing',
```
timdeschryver, line 22: Should we rename this rule to `selectors-start-with-select` ?
The description and the message could be `The variable name of a selector should start with "select", for example "selectThing"`?

src/rules/store/select-in-selector-names.ts
rafaelss95: To complement this comment from @timdeschryver, it worths to say that `/^${memorizedSelectorTypes}$/` and `/^${selectorsNames}$/` don't work the way you might be imagining... if you do it like this, you'll get the words `memorizedSelectorTypes` and `selectorsNames` instead of the computed variables.

When you want to interpolate a variable into a `RegExp` you have to do this instead:

```ts
RegExp(`^${memorizedSelectorTypes}$`)
```

and 

```ts
RegExp(\`^${selectorsNames}$`)
```

src/rules/store/select-in-selector-names.ts
rafaelss95: To be honest, you can just write a selector to avoid both the iteration mentioned above and conditionals:

```ts
create(context) {
  return {
    'VariableDeclarator[id.name=/^(?!select).+/]:matches(:has(TSTypeAnnotation[typeAnnotation.typeName.name=/^MemoizedSelector(Props$|$)/]), :has(CallExpression[callee.name=/^(createSelector|createFeatureSelector)$/]))'({
      id,
    }: TSESTree.VariableDeclarator) {
      context.report({ node: id, messageId })
    },
  }
},
```

timdeschryver: And here is our query wizard 🪄

prince-chrismc: mind-blown

prince-chrismc:
I've applied all your suggestions and my tests passed locally. 🤞 Should be good to go.

Thanks for the help getting this rule added. I am really happy I learnt a lot! (which means there's a lot left 🤣)

And on that note, after 6 months learning Flux/NgRx with my team, these rules are a huge help... They did not make much sense on day 1 but it's greatly appreciate 💟

src/rules/store/selectors-start-with-select.ts
```diff
@@ -0,0 +1,36 @@
...
+    },
+    schema: [],
+    messages: {
+      [messageId]: 'The variable name of a selector should start with "select"',
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    return {
+      'VariableDeclarator[id.name=/^(?!select).+/]:matches(:has(TSTypeAnnotation[typeAnnotation.typeName.name=/^MemoizedSelector(Props$|$)/]), :has(CallExpression[callee.name=/^(createSelector|createFeatureSelector)$/]))'({
```
rafaelss95, line 29: After a second look to this, I noticed two things here:

1. What should happen if one uses, for example, `select` or `selectxx` words? I'm asking because the current pattern treats this as valid... **but** shouldn't we ensure there's an uppercase after the 'select' word? Maybe it could impact the rule name too, as start with !== prefixed with, no? Anyway, let me know your opinions about this (it'd be nice to have test cases for this).

2. It should be `MemoizedSelector(WithProps$|$)`, not `MemoizedSelector(Props$|$)` (it's missing the 'With'). However, is it something we should care about? It's currently **deprecated** as documented in https://ngrx.io/api/store/MemoizedSelectorWithProps. If we decide to care about it, it'd be nice to have a test case for this then.

cc @timdeschryver

---

Here's the version that requires 'select[A-Z]' pattern and **with** `MemoizedSelectorWithProps`:

 ```suggestion
       'VariableDeclarator[id.name!=/^(^select)[A-Z][a-zA-Z]+$/]:matches(:has(TSTypeAnnotation[typeAnnotation.typeName.name=/^MemoizedSelector(WithProps$|$)/]), :has(CallExpression[callee.name=/^(createSelector|createFeatureSelector)$/]))'({
```

And here's the version that requires 'select[A-Z]' pattern and without `MemoizedSelectorWithProps`:

 ```suggestion
       'VariableDeclarator[id.name!=/^(^select)[A-Z][a-zA-Z]+$/]:matches(:has(TSTypeAnnotation[typeAnnotation.typeName.name=/^MemoizedSelector$/]), :has(CallExpression[callee.name=/^(createSelector|createFeatureSelector)$/]))'({
```

tests/rules/selectors-start-with-select.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      counter: number;
+    }
+
+    export interface AppState {
+      feature: FeatureState;
+    }
+
+    export const selectFeature = createFeatureSelector<AppState, FeatureState>(featureKey);`,
+  ],
+  invalid: [
```
rafaelss95, line 36: What about adding a case for `createFeatureSelector` too?

tests/rules/selectors-start-with-select.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  messageId,
+} from '../../src/rules/store/selectors-start-with-select'
+import { ruleTester } from '../utils'
+
+ruleTester().run(path.parse(__filename).name, rule, {
+  valid: [
+    `
+    interface AppState {
+      feature: FeatureState;
+    }
```
rafaelss95, line 14: Since this rule doesn't deal with type checker, coudn't we just drop this and other interfaces in this file to avoid unnecessary distractions?

rafaelss95 (commented):
Great work, @prince-chrismc 🚀 I've submitted a review about typos and one point to discuss + tests. PTAL

docs/rules/selectors-start-with-select.md
```diff
@@ -0,0 +1,30 @@
+# Selectors State with Select
+
+> The variable name of a selector should start with "select", for example "selectThing"
```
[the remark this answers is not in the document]
timdeschryver: The suggestion gets a 👍 from me

src/rules/store/selectors-start-with-select.ts
timdeschryver: I think we should do 1 (the casing) because typos are quickly made 😅
I wouldn't bother with 2 (with props), and just leave it as is 

The second suggested change seems good to me

tests/rules/selectors-start-with-select.test.ts
timdeschryver: I agree

timdeschryver (commented):
Thanks for your work @prince-chrismc and thanks for the good review @rafaelss95 🏆

src/rules/store/selectors-start-with-select.ts
rafaelss95: Nice, @timdeschryver. In this case should we still call it `selectors-start-with-select` or would it be better to call it `selectors-prefix-with-select`?

@prince-chrismc I just noticed that you opted for maintaining `MemoizedSelectorWithProps`... so in this case, it'd be good to have examples in `.md` and test cases (a success and a fail case should be good). Also, a example + test case for the word "select" 😆

docs/rules/selectors-start-with-select.md
```diff
@@ -0,0 +1,34 @@
...
+> The selector should start with "select", for example "selectThing"
+
+It's recommended prefixing selector function names with the word `select` combined with a description of the value being selected.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```ts
+// ⚠ Usage a selector without any prefix
```
rafaelss95, line 12:
```suggestion
// ⚠ Usage of a selector without any prefix
```

src/rules/store/selectors-start-with-select.ts
prince-chrismc: That was not intentional, however we still use props and this rule does not cover the `createSelectorFactory` which I believe is the replacement?

docs/rules/selectors-start-with-select.md
```diff
@@ -0,0 +1,34 @@
...
+const selectfeature = createFeatureSelector<AppState, FeatureState>(featureKey)
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+export const selectFeature = createSelector((state: AppState) => state.feature)
+export const selectFeature: MemoizedSelector<any, any> = (state: AppState) =>
+  state.feature
+const selectFeature = createFeatureSelector<AppState, FeatureState>(featureKey)
```
timdeschryver, line 29: This is also a lint rule ;) 

```suggestion
const selectFeature = createFeatureSelector<FeatureState>(featureKey)
```

tests/rules/selectors-start-with-select.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ruleTester } from '../utils'
+
+ruleTester().run(path.parse(__filename).name, rule, {
+  valid: [
+    `
+    export const selectFeature:MemoizedSelector<any, any> = (state: AppState) => state.feature`,
+    `
+    export const selectFeature = createSelector((state: AppState) => state.feature)`,
+    `
+    export const selectFeature = createFeatureSelector<AppState, FeatureState>(featureKey);`,
```
timdeschryver, line 16:
```suggestion
    `
    export const selectFeature = createFeatureSelector<FeatureState>(featureKey);`,
    `
    export const selectFeature = createFeatureSelector<AppState, FeatureState>(featureKey);`,
```

timdeschryver:
@all-contributors please add @prince-chrismc for code, idea

src/rules/store/selectors-start-with-select.ts
timdeschryver: I think `prefix-selectors-with-select` would be the best then if both of you agree.
The replacement is a selector that returns a selector, e.g.

```ts
const selectThingy = (id: string) => createSelector(selectThings, things => things[id]);
```

I think for now, this rule covers like 90% of the cases and we can merge this in if we agree on the name.
Later, we can follow-up with a PR to address the factory selectors.

timdeschryver:
>  And on that note, after 6 months learning Flux/NgRx with my team, these rules are a huge help... They did not make much sense on day 1 but it's greatly appreciate

@prince-chrismc if something wasn't clear why some rules exist, feel free to open PRs to add more info to the docs or to open issues with questions.

src/rules/store/selectors-start-with-select.ts
rafaelss95: I agree! 👍

timdeschryver (approved):
LGTM, thanks again!

rafaelss95 (approved):
LGTM 🚀

PR #171, opened 2021-08-20
fix(avoid-cyclic-effects): warn if ofType has same type as source

rafaelss95:
I just noticed some unused dependencies listed as `devDependencies`. There are:

- `@angular/core`
- `@ngrx/component-store`
- <s>`@ngrx/effects`</s> -> Used in `avoid-cyclic-effects`
- <s>`@ngrx/store`</s> -> Used in `avoid-cyclic-effects`
- `rxjs`

This PR aims to remove them and taking advantage I was looking at this, I decided to update all the other dependencies to the most suitable version (this also solves some high severity vulnerabilities encountered on installation).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,6 @@ module.exports = {
   extends: [
     'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended',
-    'prettier/@typescript-eslint',
     'plugin:prettier/recommended',
   ],
   parserOptions: {
--- a/src/rules/effects/avoid-cyclic-effects.ts
+++ b/src/rules/effects/avoid-cyclic-effects.ts
@@ -103,7 +103,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
         return memberActionTypes
       }
       const symbol = typeChecker.getPropertyOfType(type, 'type')
-      if (!symbol) {
+      if (!symbol?.valueDeclaration) {
         return []
       }
       const actionType = typeChecker.getTypeOfSymbolAtLocation(
--- a/src/rules/effects/avoid-cyclic-effects.ts
+++ b/src/rules/effects/avoid-cyclic-effects.ts
@@ -65,9 +65,8 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
       if (!isTypeReference(operatorReturnType)) {
         return
       }
-      const [operatorElementType] = typeChecker.getTypeArguments(
-        operatorReturnType,
-      )
+      const [operatorElementType] =
+        typeChecker.getTypeArguments(operatorReturnType)
       if (!operatorElementType) {
         return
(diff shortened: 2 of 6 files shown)


.eslintrc.js
```diff
@@ -4,7 +4,6 @@ module.exports = {
   extends: [
     'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended',
-    'prettier/@typescript-eslint',
```
rafaelss95, line 7: https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#version-800-2021-02-21

timdeschryver (commented):
Most (maybe all) of the listed dependencies are needed to infer the correct types in our tests. If I'm not mistaken...

rafaelss95:
> Most (maybe all) of the listed dependencies are needed to interfer the correct types in our tests.

You're correct, I just noticed that `@ngrx/effects` and `@ngrx/store` are used to test `avoid-cyclic-effects`. I have updated it right now, PTAL :)

src/rules/effects/avoid-cyclic-effects.ts
```diff
@@ -104,7 +103,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
         return memberActionTypes
       }
       const symbol = typeChecker.getPropertyOfType(type, 'type')
-      if (!symbol) {
+      if (!symbol?.valueDeclaration) {
```
timdeschryver, line 106: I suppose this was the fix for the "todo test"?
If so, I suppose we should commit this as a fix, so it releases a new version.

rafaelss95: Not really. This was changed because now [`Symbol.valueDeclaration`](https://github.com/microsoft/TypeScript/blob/e00722f262f8a84c27f01e2b6a41eecb87c4afd4/src/compiler/types.ts#L4833) is possibly `undefined` and `TypeChecker.getTypeOfSymbolAtLocation` only accepts a `Node`, not `Node | undefined`.

Note that it was reported in https://github.com/microsoft/TypeScript/issues/24706 and fixed in https://github.com/microsoft/TypeScript/pull/43033). Also, this was also fixed in the [no-cyclic-action.ts](https://github.com/cartant/eslint-plugin-rxjs/blob/main/source/rules/no-cyclic-action.ts) in this [commit](https://github.com/cartant/eslint-plugin-rxjs/commit/fee48bceda7d214d3ef52234428620d302e5ee7f)

---

That said, I must confess that I didn't do anything to fix that test case. I was simply testing all the tests to see if everything was working as before and I realized that one case was failing, so I came across to that comment and I simply moved it to invalid section 😆.

timdeschryver: 😅 that's why I couldn't figure out what the actual fix was.
Thanks for the answer!

timdeschryver (approved):

PR #176, opened 2021-08-30
fix(no-store-subscription): only handle store subscriptions

### LINKED ISSUE #175: False positive for no-store-subscription
## Behavior

It highlights not only subscriptions to the store, but just any subscription if a store is involved.

## Code snippet

```
    this.formCtrlOrg.valueChanges.pipe(takeUntil(this.drop)).subscribe((orgName) => {
      this.store.dispatch(UserPageActions.checkOrgNameInput({orgName}));
    });
```

rafaelss95:
Fixes #175.

### DIFF
--- a/src/rules/store/no-store-subscription.ts
+++ b/src/rules/store/no-store-subscription.ts
@@ -2,7 +2,7 @@ import path from 'path'
 import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
 import { findNgRxStoreName } from '../../utils/helper-functions/index'
 
-import { docsUrl, storeExpression } from '../../utils'
+import { docsUrl } from '../../utils'
 
 export const messageId = 'noStoreSubscription'
 export type MessageIds = typeof messageId
--- a/src/rules/store/no-store-subscription.ts
+++ b/src/rules/store/no-store-subscription.ts
@@ -31,9 +31,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
     if (!storeName) return {}
 
     return {
-      [`CallExpression:has(${storeExpression(
-        storeName,
-      )}) > MemberExpression > Identifier[name='subscribe']`](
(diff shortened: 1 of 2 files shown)


src/rules/store/no-store-subscription.ts
```diff
@@ -31,9 +31,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
     if (!storeName) return {}
 
     return {
-      [`CallExpression:has(${storeExpression(
-        storeName,
-      )}) > MemberExpression > Identifier[name='subscribe']`](
+      [`CallExpression:matches([callee.object.callee.object.name=${storeName}], [callee.object.callee.object.object.type='ThisExpression'][callee.object.callee.object.property.name=${storeName}]) > MemberExpression > Identifier[name='subscribe']`](
```
rafaelss95, line 34: Unfortunately I don't think we can reuse the `storeExpression` function anymore, but let me know if you have any suggestions.

tests/rules/no-store-subscription.test.ts
```diff
@@ -11,17 +11,45 @@ ruleTester().run(path.parse(__filename).name, rule, {
       @Component()
```
rafaelss95, line 11: Do you know how could I test a single file with `uvu`? Also, is there any way to check what test file is not behaving as expected? Currently I got this:

Captura de Tela 2021-08-30 às 10 57 54

timdeschryver: Yikes, a single test can be run with the following command.
Sadly it doesn't support a glob.

```
node -r ts-node/register ./tests/rules/no-store-subscription.test.ts
```

For the AssertionError, I think we'll have to wrap the testrunner to also print the file 🤔
I'll try to work on this soon.

src/rules/store/no-store-subscription.ts
timdeschryver: I think you're right.
Does this mean that we can remove `storeExpression` ?

timdeschryver (approved):
LGTM, let's get this merged! 🚀

PR #177, opened 2021-08-30
fix(avoid-mapping-selectors): report within `pipe` and `select`

### LINKED ISSUE #174: False positive with combineLatest (ngrx/avoid-mapping-selectors)
## Behavior

When combining store with another observable and map the linter tells me to use selectors

## Code snippet

```ts
combineLatest([this.store.select(selectAuthorizations), this.hasAuthorization$]).pipe(map((val) => !isEmpty(intersection(val))))
```

[x] No

rafaelss95:
Fixes #174.

### DIFF
--- a/src/rules/store/avoid-mapping-selectors.ts
+++ b/src/rules/store/avoid-mapping-selectors.ts
@@ -1,7 +1,12 @@
-import path from 'path'
 import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
+import path from 'path'
 
-import { docsUrl, findNgRxStoreName, storeSelect, storePipe } from '../../utils'
+import {
+  docsUrl,
+  findNgRxStoreName,
+  storeExpressionCallable,
+  storePipe,
+} from '../../utils'
 
 export const messageId = 'avoidMapppingSelectors'
 export type MessageIds = typeof messageId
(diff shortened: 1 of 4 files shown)


src/rules/store/avoid-mapping-selectors.ts
```diff
@@ -29,16 +34,15 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+    )}:has(CallExpression[callee.name='select'] ~ CallExpression[callee.name='map'])`
+    const selectSelector = `${storeExpressionCallable(
       storeName,
-    )}:has(CallExpression[callee.name="select"])`
-    const pipeExpressions = [mapInSelect, mapInPipe].join(', ')
+    )}[callee.object.callee.property.name='select']`
 
     return {
-      [`:matches(${pipeExpressions}) > CallExpression[callee.name='map']`](
+      [`:matches(${selectSelector}, ${pipeWithSelectAndMapSelector}) CallExpression[callee.name='map']`](
```
rafaelss95, line 45: I think it worths to mention that the following case is not covered by this selector:

subsequent pipes

But I don't think people really use it like this in real world 🤷🏼‍♂️.

timdeschryver: Yea... we're not going to support that 😅

rafaelss95:
```suggestion
      [`:matches(${selectSelector}, ${pipeWithSelectAndMapSelector}) > CallExpression[callee.name='map']`](
```

timdeschryver (approved):

PR #178, opened 2021-08-30
style: format files with prettier

rafaelss95:


src/rules/effects/use-effects-lifecycle-interface.ts
```diff
@@ -59,10 +59,8 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
 
         if (interfaces.includes(interfaceName)) return
 
-        const {
-          implementsNodeReplace,
-          implementsTextReplace,
-        } = getImplementsSchemaFixer(classDeclaration, interfaceName)
+        const { implementsNodeReplace, implementsTextReplace } =
+          getImplementsSchemaFixer(classDeclaration, interfaceName)
```
rafaelss95, line 63: I was seeing the following report by `prettier` (here and in /guards):

Prettier report

src/utils/helper-functions/guards.ts
```diff
@@ -1,9 +1,10 @@
 import { AST_NODE_TYPES, TSESTree } from '@typescript-eslint/experimental-utils'
 import ts from 'typescript'
 
-const isNodeOfType = <NodeType extends AST_NODE_TYPES>(nodeType: NodeType) => (
-  node: TSESTree.Node,
-): node is TSESTree.Node & { type: NodeType } => node.type === nodeType
+const isNodeOfType =
+  <NodeType extends AST_NODE_TYPES>(nodeType: NodeType) =>
+  (node: TSESTree.Node): node is TSESTree.Node & { type: NodeType } =>
+    node.type === nodeType
```
rafaelss95, line 7: I was seeing the following report by `prettier` (here and in /use-effects-lifecycle-interface.ts):

Prettier report

src/utils/helper-functions/guards.ts
```diff
@@ -30,6 +31,5 @@ export const isObjectExpression = isNodeOfType(AST_NODE_TYPES.ObjectExpression)
 export const isProperty = isNodeOfType(AST_NODE_TYPES.Property)
 
 export function isTypeReference(type: ts.Type): type is ts.TypeReference {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return Boolean((type as any).target)
+  return type.hasOwnProperty('target')
```
rafaelss95, line 34: Taking into consideration the fact that I was messing up this file, I changed the implementation here so we can remove the typecast.

timdeschryver (approved):
LGTM.
Do you think we need a check in our CI that verifies if prettier has run?

src/utils/helper-functions/guards.ts
timdeschryver: Good catch!

rafaelss95:
> LGTM.
> Do you think we need a check in our CI that verifies if prettier has run?

It'd be great 😄

PR #180, opened 2021-08-31
refactor(avoid-dispatching-multiple-actions-sequentially): simplify

rafaelss95:
With this we no longer need to store the grandparent as the key of `Map` (we don't even need a `Map` + looking to the parents anymore) in order to check if it's within the same `BlockStatement`.

If you agree, I can also refactor `avoid-duplicate-actions-in-reducer` (currently it's checking/reporting duplication on every `createReducer` > `Identifier`).


timdeschryver (commented):
I think this is a smart solution, but I don't think it works for the following case.
This case, also doesn't catch violations with the current implementation though 😅.



```
pingPong() {
		  // this creates another "block"
          if(true){
            this.customName.dispatch({ type: 'PING' })
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [${messageId}]
          }
          this.customName.dispatch({ type: 'PONG' })
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [${messageId}]
        }
```

rafaelss95:
Hmm, this case is really complicated... do you think is it something we should have to worry about? If so, we would have to check if the conditional evaluates to true and most cases are not as trivial as `if (true) { ... }`, so it would involve type-checking probably 😅

timdeschryver:
I'm thinking about it, I was thinking we could scan a `MethodDefinition` for multiple dispatches, but that's also incorrect for the following case. So the current implementation is alright for  most of the cases, which means we could use the refactored version. 

```ts
class Foo {

	doSomething() {
		if(true){ 
			dispatch()
		}else {
			dispatch()
		}
	}
}
```

timdeschryver:
Oh sorry @rafaelss95 
I was just fixing it, but you beat me to it 😅

timdeschryver:
It was a good test to verify our lint check works though 😅

timdeschryver (approved):

PR #181, opened 2021-08-31
ci: lint during ci

timdeschryver:


package.json
```diff
@@ -34,12 +34,15 @@
   ],
   "scripts": {
     "test": "uvu -r ts-node/register tests",
+    "test:one": "node -r ts-node/register",
```
timdeschryver, line 37: ideally, I wanted to just pass the rule name and it would run it...
but I couldn't figure out how to make it cross-platform without creating a custom script for it, might do that later 😅
for now, you can pass the path to the test

rafaelss95: Nice! That works for me 🥇

rafaelss95 (approved):
LGTM!

PR #182, opened 2021-08-31
style: enforce consistent type-import

rafaelss95:
Currently we have incosistencies related to `type-only imports` vs `imports`. Using the rule [**`@typescript-eslint/consistent-type-imports`**](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md) we can now keep it consistent.

### DIFF
--- a/src/rules/component-store/updater-explicit-return-type.ts
+++ b/src/rules/component-store/updater-explicit-return-type.ts
@@ -1,5 +1,6 @@
+import type { TSESTree } from '@typescript-eslint/experimental-utils'
+import { ESLintUtils } from '@typescript-eslint/experimental-utils'
 import path from 'path'
-import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
 
 import { docsUrl, findNgRxComponentStoreName } from '../../utils'
 
--- a/src/rules/component-store/updater-explicit-return-type.ts
+++ b/src/rules/component-store/updater-explicit-return-type.ts
@@ -1,7 +1,6 @@
 import type { TSESTree } from '@typescript-eslint/experimental-utils'
 import { ESLintUtils } from '@typescript-eslint/experimental-utils'
 import path from 'path'
-
 import { docsUrl, findNgRxComponentStoreName } from '../../utils'
 
 export const messageId = 'updaterExplicitReturnType'
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
     extraFileExtensions: ['.json'],
   },
   rules: {
+    '@typescript-eslint/consistent-type-imports': 'error',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
     '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,6 +31,7 @@
         "husky": "^7.0.1",
         "lint-staged": "^11.1.2",
         "prettier": "^2.3.2",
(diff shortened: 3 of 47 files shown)


src/rules/component-store/updater-explicit-return-type.ts
```diff
@@ -1,5 +1,6 @@
+import type { TSESTree } from '@typescript-eslint/experimental-utils'
+import { ESLintUtils } from '@typescript-eslint/experimental-utils'
 import path from 'path'
-import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
 
 import { docsUrl, findNgRxComponentStoreName } from '../../utils'
```
rafaelss95, line 5: Question: is this blank line between the 3rd. imports and the project imports on purpose? I ask because I saw it is present in some files and not others... To be honest, I think not having this line is much better since I can just save and the VSCode (in my case) organizes everything automatically.

timdeschryver: I don't have a strong opinion on this, if vscode can sort it automatically we should just respect that and remove the line. In some projects this is a convention to differentiate libs with imports from your own code base. So if you want, you can remove the extra line before this gets merged.

rafaelss95: Done! To make this change in all files at once, I've installed the extension [folder-source-actions](https://marketplace.visualstudio.com/items?itemName=bierner.folder-source-actions), but normally(

```json
    "editor.codeActionsOnSave": {
      "source.organizeImports": true
    }
```

) on `settings.json` is sufficient for me 😄

timdeschryver: Good to know that there's a plugin for it!
I think we could also enforce this with the [sort-imports rule](https://eslint.org/docs/rules/sort-imports)?

rafaelss95: Hmmmmm, this rule acts a little different... it doesn't just allow us to sort the imports alphabetically, no matter what type, or rather `memberSyntaxSortOrder`, like the VSCode command.

What I see people doing, if they want to leave it in lint too, is to use a 3rd. plugin like [**`eslint-plugin-import/order.md`**](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md) or a simpler version like [**`simple-import-sort/imports`**](https://github.com/lydell/eslint-plugin-simple-import-sort/issues/65)... however, these still are not compatible with simply _organize my imports_, as it still requires us to choose a `groups` configuration (or to use the default one).

Looking for way to truly behave like the VSCode `organizeImports`, I found this [issue](https://github.com/lydell/eslint-plugin-simple-import-sort/issues/65) and found two interesting approaches to go with [here](https://github.com/lydell/eslint-plugin-simple-import-sort/issues/65#issuecomment-763006464): either enforce this using prettier through [**`prettier-plugin-organize-imports`**](https://github.com/simonhaenisch/prettier-plugin-organize-imports) or [**`organize-imports-cli`**](https://github.com/thorn0/organize-imports-cli). The first option (`prettier-plugin-organize-imports`) seems to be more popular, but it's up to you whether to add this here (IMHO it's a nice thing to have).

timdeschryver: Feel free to add it if you want @rafaelss95 👍

rafaelss95: Done!

timdeschryver (approved):

timdeschryver:
Thanks for keeping this repo clean 🧹

PR #184, opened 2021-09-03
fix: ability to handle multiple imports

rafaelss95:
Currently, the `utils/findBy*` isn't able to handle multiple imports... so when I have something like this:

```ts
import type { Something } from '@ngrx/store'
```

...the `findNgRxStoreName` returns null as it stops at the first `ImportDeclaration`.

By the way, it was a real situation in an application I was working on, some rules weren't behaving correctly as the imports, like here, are split between type-only and conventional imports 😅 

I've added a test for one rule of each package, LMK if it's sufficient :)

### DIFF
--- a/src/utils/helper-functions/utils.ts
+++ b/src/utils/helper-functions/utils.ts
@@ -11,7 +11,9 @@ import {
 } from './guards'
 
 export const MODULE_PATHS = {
+  componentStore: '@ngrx/component-store',
   effects: '@ngrx/effects',
+  store: '@ngrx/store',
 }
 
 export function findClassDeclarationNode(
@@ -92,20 +94,24 @@ export function getDecoratorArgument({ expression }: TSESTree.Decorator) {
     : undefined
 }
 
-export function findNgRxStoreName(
+function findCorrespondingNameBy(
   context: TSESLint.RuleContext<string, readonly unknown[]>,
+  moduleName: string,
+  importedName: string,
 ): string | undefined {
   const { ast } = context.getSourceCode()
-  const storeImportSpecifier = findImportDeclarationNode(ast, '@ngrx/store')
+  const importDeclarations = getImportDeclarations(ast, moduleName) ?? []
+  const { importSpecifier } =
+    getImportDeclarationSpecifier(importDeclarations, importedName) ?? {}
 
-  if (!storeImportSpecifier) return undefined
+  if (!importSpecifier) {
+    return undefined
+  }
 
-  const variables = context.getDeclaredVariables(storeImportSpecifier)
-  const storeVariable = variables.find((v) => v.name === 'Store')
+  const variables = context.getDeclaredVariables(importSpecifier)
+  const typedVariable = variables.find(({ name }) => name === importedName)
 
-  if (!storeVariable) return undefined
-
-  return storeVariable.references
+  return typedVariable?.references
     .map(({ identifier: { parent } }) => {
       if (
         parent &&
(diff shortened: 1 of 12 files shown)


src/utils/helper-functions/utils.ts
```diff
@@ -11,63 +10,100 @@ import {
...
   effects: '@ngrx/effects',
+  store: '@ngrx/store',
 }
 
-export function findClassDeclarationNode(
-  node: TSESTree.Node,
-): TSESTree.ClassDeclaration | null {
-  if (isClassDeclaration(node)) {
-    return node
+export function getNearestUpperNodeFrom<T extends TSESTree.Node>(
```
rafaelss95, line 18: Instead of having a function for getting a specific type, IMHO we can have a function to lookup given a predicate over parents.

src/utils/helper-functions/utils.ts
```diff
@@ -11,63 +10,100 @@ import {
...
+    parent = parent.parent
   }
-  if (!node.parent) return null
-  return findClassDeclarationNode(node.parent)
+
+  return undefined
 }
 
-export function findImportDeclarationNode(
+export function getImportDeclarationSpecifier(
```
rafaelss95, line 33: These functions were copied from my original implementation at https://github.com/angular-eslint/angular-eslint/blob/a446e8ff521725d354dc23242c4ad23bc52c9681/packages/eslint-plugin/src/utils/utils.ts 👀

src/utils/helper-functions/utils.ts
```diff
@@ -11,63 +10,100 @@ import {
...
+
+  return parentNode?.body.filter(
+    (node): node is TSESTree.ImportDeclaration =>
+      isImportDeclaration(node) && node.source.value === moduleName,
+  )
 }
 
-export function getConditionalImportFix(
-  fixer: TSESLint.RuleFixer,
+export function getImportAddFix(
```
rafaelss95, line 71: Just noticed that this implementation is problematic for default imports, namespace imports and type-only imports, as it could potentially introduce a break code, i.e.:

```ts
import * as ngrx from '@ngrx/store'

// would become something like

import * as ngrx, Effect from '@ngrx/store'

```

...let me try something better (though it's broken in the current version also).

rafaelss95: It should be good now.

tests/rules/use-effects-lifecycle-interface.test.ts
```diff
@@ -122,6 +123,7 @@ ruleTester().run(path.parse(__filename).name, rule, {
       {
         output: stripIndent`
           import { OnInitEffects } from '@ngrx/effects';
+          import * as ngrx from '@ngrx/effects'
```
timdeschryver, line 126: The extra imports are here just to test that we correctly fix the import, right?
If that's that case and I'm being honest, then I do find it a bit confusing.
The reason is that I like to have the cases in here, as compact as possible.
But I also get that we want to test that it works 😅.

rafaelss95: Yes, that's it. I'm also part of the team that likes compact testing 😄

---

Taking advantage of the fact that we're talking about testing... I was looking at how `@typescript-eslint` does the tests and I found an interesting approach there: they don't mess with formatting literals, like these for example:

Captura de Tela 2021-09-04 às 01 00 06

Source: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/tests/rules/no-unsafe-argument.test.ts

Do you think we could follow something similar here? I ask because I see that we have some formatting inconsistencies between the tests, which could be remedied this way. An example:

Captura de Tela 2021-09-04 às 01 03 17

timdeschryver: Sure we can do that, I stole the `stripIndent` from the rjxs linter 😅
I'm don't have a big opinion, but I would say let's try to keep things consistent with the Angular one.
I'll review the other PRs on Monday.
Have a great weekend! ✨

timdeschryver (approved):

PR #185, opened 2021-09-04
style: standarize `docs#description` and `messageId`

rafaelss95:
A simple PR that adds the missing punctuations to `docs#description` + `messageId` 😆


rafaelss95:
Thinking more about this... to make sure we keep it consistent, I was looking for packages for this and I found https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin, which contains the rules [**`report-message-format`**](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/report-message-format.md) and [**`require-meta-docs-description`**](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-docs-description.md), however we can't simply use this package as its implementation doesn't cover our way of creating rules.

I was thinking, what about creating an internal package for this like, for example, https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin-internal?

timdeschryver (approved):
You mean that we need a rule that verifies the format of the description and the messages? e.g. that it starts with an uppercase, and ends with a dot?
AFAIK, we could create internal rules for this repo and add them to our eslint config.

rafaelss95:
Yes, that's what I mean :)

PR #186, opened 2021-09-04
fix(avoid-duplicate-actions-in-reducer): simplify

rafaelss95:
Currently it's checking/reporting duplication on every `createReducer > Identifier`). With the new implementation we collect all the possible duplicate nodes and report it on the `createReducer:exit` as mentioned in #180.

This unintentionally also fixes a possible false positive that is having different reducers controlling actions with the same names :stuck_out_tongue:


timdeschryver (approved):
I'm going to commit this as a fix, so the accidental fix is also released 🚀

PR #187, opened 2021-09-04
refactor: use `as const` for selectors

rafaelss95:
**Before**:

Captura de Tela 2021-09-04 às 02 19 23

**After**:

Captura de Tela 2021-09-04 às 02 19 45

IMHO, it's much nicer to see what is the content of the select by just hovering it, instead of going to that file.

### DIFF
--- a/src/utils/selectors/index.ts
+++ b/src/utils/selectors/index.ts
@@ -1,12 +1,15 @@
 export const effectCreator = `ClassProperty[value.callee.name='createEffect']`
 
 export const effectDecorator = `Decorator[expression.callee.name='Effect']`
-export const classPropertyWithEffectDecorator = `ClassDeclaration > ClassBody > ClassProperty > ${effectDecorator}`
+export const classPropertyWithEffectDecorator =
+  `ClassDeclaration > ClassBody > ClassProperty > ${effectDecorator}` as const
 
 export const actionCreator = `CallExpression[callee.name='createAction']`
-export const actionCreatorWithLiteral = `${actionCreator}[arguments.0.type='Literal']`
-export const actionCreatorProps = `${actionCreator} CallExpression`
-export const actionCreatorPropsComputed = `${actionCreatorProps} > TSTypeParameterInstantiation > :matches(TSTypeReference[typeName.name!='Readonly'], [type=/^TS(.*)(Keyword|Type)$/])`
+export const actionCreatorWithLiteral =
+  `${actionCreator}[arguments.0.type='Literal']` as const
+export const actionCreatorProps = `${actionCreator} CallExpression` as const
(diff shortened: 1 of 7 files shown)


src/utils/selectors/index.ts
```diff
@@ -18,9 +21,11 @@ export const typedStore = `MethodDefinition[kind='constructor'] Identifier>TSTyp
...
-export const ngModuleProviders = `${ngModuleDecorator} ObjectExpression Property[key.name='providers'] > ArrayExpression Identifier`
+export const ngModuleProviders =
+  `${ngModuleDecorator} ObjectExpression Property[key.name='providers'] > ArrayExpression Identifier` as const
 
-export const ngModuleImports = `${ngModuleDecorator} ObjectExpression Property[key.name='imports'] > ArrayExpression CallExpression[callee.object.name='EffectsModule'][callee.property.name=/forRoot|forFeature/] ArrayExpression > Identifier`
+export const ngModuleImports =
+  `${ngModuleDecorator} ObjectExpression Property[key.name='imports'] > ArrayExpression CallExpression[callee.object.name='EffectsModule'][callee.property.name=/forRoot|forFeature/] ArrayExpression > Identifier` as const
 
 export const actionDispatch = (storeName: string) =>
   `ExpressionStatement > CallExpression:matches([callee.object.name='${storeName}'][callee.property.name='dispatch'], [callee.object.object.type='ThisExpression'][callee.object.property.name='${storeName}'][callee.property.name='dispatch'])`
```
rafaelss95, line 31: Although we could do for functions/selectors that deal with `storeName`, it wouldn't be used much since the storeName is fully dynamic.

Not to mention that, honestly, it would be quite verbose like this:

```suggestion
export const actionDispatch = <TStore extends string>(
  storeName: TStore,
): `ExpressionStatement > CallExpression:matches([callee.object.name='${TStore}'][callee.property.name='dispatch'], [callee.object.object.type='ThisExpression'][callee.object.property.name='${TStore}'][callee.property.name='dispatch'])` =>
  `ExpressionStatement > CallExpression:matches([callee.object.name='${storeName}'][callee.property.name='dispatch'], [callee.object.object.type='ThisExpression'][callee.object.property.name='${storeName}'][callee.property.name='dispatch'])`
```

timdeschryver: Seems like a good call 😅

timdeschryver (approved):
🤩 Nice!
I always forget this trick.

src/utils/selectors/index.ts
rafaelss95: I just noticed that we could add `as const` without the need of explicit return type, so instead of `string` we get this:

as const

PR #188, opened 2021-09-04
fix: handle literals properly

rafaelss95:
Currently the rules `no-effects-in-providers` and `no-reducer-in-key-names` aren't capable of looking at `@NgModule` metadata if it's a computed literal or a template literal.

### DIFF
--- a/src/utils/selectors/index.ts
+++ b/src/utils/selectors/index.ts
@@ -14,36 +14,54 @@ export const actionCreatorPropsComputed =
 export const constructorExit = `MethodDefinition[kind='constructor']:exit`
 
 export const dispatchInEffects = (storeName: string) =>
-  `ClassProperty > CallExpression:has(Identifier[name='createEffect']) CallExpression > MemberExpression:has(Identifier[name='dispatch']):has(MemberExpression > Identifier[name='${storeName}'])`
+  `ClassProperty > CallExpression:has(Identifier[name='createEffect']) CallExpression > MemberExpression:has(Identifier[name='dispatch']):has(MemberExpression > Identifier[name='${storeName}'])` as const
 
 export const injectedStore = `MethodDefinition[kind='constructor'] Identifier[typeAnnotation.typeAnnotation.typeName.name='Store']`
 export const typedStore = `MethodDefinition[kind='constructor'] Identifier>TSTypeAnnotation>TSTypeReference[typeName.name='Store'][typeParameters.params]`
 
 export const ngModuleDecorator = `ClassDeclaration > Decorator > CallExpression[callee.name='NgModule']`
 
-export const ngModuleProviders =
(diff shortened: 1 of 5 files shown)


src/utils/selectors/index.ts
```diff
@@ -18,9 +18,29 @@ export const typedStore = `MethodDefinition[kind='constructor'] Identifier>TSTyp
...
+  `${ngModuleImports} CallExpression[callee.object.name='EffectsModule'][callee.property.name=/^for(Root|Feature)$/] ArrayExpression > Identifier` as const
+
+export const effectsInNgModuleProviders =
+  `${ngModuleProviders} Identifier` as const
+
+export function metadataProperty(key: RegExp): string
+export function metadataProperty<TKey extends string>(
+  key: TKey,
+): `Property:matches([key.name=${TKey}][computed=false], [key.value=${TKey}], [key.quasis.0.value.raw=${TKey}])`
+export function metadataProperty(key: RegExp | string): string {
```
rafaelss95, line 41: I noticed that you seem to prefer `arrow function` to `function`, however as it needs to be overloaded, I used `function` here.

timdeschryver (commented):
I do like the refactored version, but do you think that we need to keep the "complexity" of string literals as property names? I'm asking because I've never seen them being used (in these places) in a project.

rafaelss95:
> I do like the refactored version, but do you think that we need to keep the "complexity" of string literals as property names? I'm asking because I've never seen them being used (in these places) in a project.

To be quite honest, I'm most interested in reporting keys with quotes as `{ 'imports': [] }` and to **not** report computed keys like `[imports]: []` effectivelly, the rest comes with as a bonus 😅. So, in order to maintain the behavior consistent throught the rules I created this selector function so we can reuse it anywhere we need this kind of verification.

For example, in the `no-effects-in-providers` we have this behavior:

```ts
@NgModule({
  imports: [...] // ✅ 
  'imports': [...] // ❌ (not being catch by the actual selector)
  [imports]: [...] // ❌ (currently a false positive, because we can't assume it's a truly 'imports' since it's a computed variable),
  ['imports']: [...] // ❌ (ok, this a uncommon case, but since we should support the two cases above I don't think it hurts to leave it here too :)
  [`imports`]: [...] // ❌ same as above
})
export class CustomModule {}
```

and the `no-reducer-in-key-names`, this:

```ts
@NgModule({
  imports: [...] // ✅ 
  'imports': [...] // ✅
  [imports]: [...] // ❌ (currently a false positive, because we can't assume it's a truly 'imports' since it's a computed variable),
  ['imports']: [...] // ❌ (ok, this a uncommon case, but since we should support the case above I don't think it hurts to leave it here too :)
  [`imports`]: [...] // ❌ same as above
})
export class CustomModule {}
```

timdeschryver (approved):

Issue #179, opened 2021-08-31
Add more fixers/suggestions


rafaelss95:
## Information about the new behavior
Thinking about a better DX, we could add fixers/suggestions for some more rules.

| Rule                               | Fixer                                                | Suggestions                                                               | PR <a href="">#<a> |
|------------------------------------|------------------------------------------------------|---------------------------------------------------------------------------|--------------------|
| `avoid-duplicate-actions-in-reducer`           | ❌                                                    | ✅                                                                         |  #209                   |
| `no-dispatch-in-effects`           | ❌                                                    | ✅                                                                         | #211                   |
| `no-effect-decorator-and-creator`  | ✅                                                    | ⚠️ (if the `@Effect` contains arguments, the fixer would not be safe.)    | #191               |
| `no-effects-in-providers`          | Probably ✅ (can this cause any unexpected behavior?) | ❌                                                                         | #203                   |
| `no-multiple-global-stores`        | ❌                                                    | ✅                                                                         | #198                   |
| `no-reducer-in-key-names`          | ❌                                                    | ✅                                                                         |  #200                   |
| `on-function-explicit-return-type` | ❌                                                    | ⚠️  (we can suggest `State` as it's a common convention in reducer files) | #199                   |
| `prefix-selectors-with-select` | ❌                                                    | ✅                                                                         | #205                  |
| `select-style`                     | ✅                                                    | ❌                                                                         | #189               |
| `use-consistent-global-store-name` | ❌                                                    | ✅                                                                         | #197                   |

Let me know your thoughts on this, @timdeschryver :)

## I would be willing to submit a PR for the docs :heart:

[x] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

timdeschryver:
I would like to know what the fixer/suggestion would do 😅
Also, the `@effect` decorator will be removed soon - so this is also a low prio for me.

Am I correct to assume that the fixer/suggestion would do the following?
no-dispatch-in-effects: remove the dispatch 
no-effects-in-providers: remove the provider
no-multiple-global-stores: remove one store, what about the references to that store?
no-reducer-in-key-names: this might be tricky, because it might impact other code (e.g. selectors) - so I would prefer to not have a fixer here
on-function-explicit-return-type: I think that `State` is used in a lot of example, but in real world applications I see a lot of "named" states, e.g. `PersonsState`. If we could infer the State interface that would be great, but I'm not sure if this is doable
[...]

PR #189, opened 2021-09-04
feat(select-style): add fixer

rafaelss95:
Part of #179.


tests/rules/select-style.test.ts
```diff
@@ -56,67 +56,115 @@ ruleTester().run(path.parse(__filename).name, rule, {
...
+          foo$ = this.store.select(selector)
+                            ~~~~~~ [${operatorSelectMessageId}]
+          constructor(private store: Store) {}
         }
-
       `,
       {
         options: [OPERATOR],
+        output: stripIndent`
+          import { select } from '@ngrx/store';
```
timdeschryver, line 159: Would it be hard to add this to the existing import?

timdeschryver (commented):
This is great, and I'm sure that it will help some codebases! 
I'm going to take a look at the implementation later, but I already have a question.

tests/rules/select-style.test.ts
rafaelss95: If my implementation is correct, it should be fine now 😅

tests/rules/select-style.test.ts
```diff
@@ -15,25 +15,25 @@ ruleTester().run(path.parse(__filename).name, rule, {
     import { Store } from '@ngrx/store'
```
rafaelss95, line 15: Something that came to my mind... since we have some rules relying on `Store` import, wouldn't it be a good idea to have a rule like `use-consistent-global-import` or something like this? Because if I do `import { Store as ngrxStore } from '@ngrx/store'` these rules won't work as expected...

timdeschryver (approved):
Looks good to me!
I'm so excited about this fix 🚀
About your question, I think it's good to keep in mind, but I also think we won't get a lot of benefit from it. Maybe we should make the store lookup mechanism smarter, so it looks for the renamed node, e.g. `ngrxStore` instead of `Store`?

rafaelss95:
> About your question, I think it's good to keep in mind, but I also think we won't get a lot of benefit from it. Maybe we should make the store lookup mechanism smarter, so it looks for the renamed node, e.g. `ngrxStore` instead of `Store`?

I just noticed that we already handle these cases properly in rules that rely on `findNgRxStoreName` here:

https://github.com/timdeschryver/eslint-plugin-ngrx/blob/c4f31b66dc81277d9d8a19ef7dd34b98e6ac3c65/src/utils/helper-functions/utils.ts#L44

Using `imported.name` is sufficient to cover this.

The only missing ones are that on rely on `typeName`:

https://github.com/timdeschryver/eslint-plugin-ngrx/blob/main/src/utils/selectors/index.ts#L19-L20

But yes, I don't think that it worths to have a rule for that.

timdeschryver:
@rafaelss95 this works like a charm!
One nit, would it be possible to remove the `select` import if we switch from the operator syntax to the method syntax?

For example:

```ts
// before
import { select, Store } from '@ngrx/store';
this.searchQuery$ = store.pipe(select(fromBooks.selectSearchQuery));

// becomes (notice that `select` is still imported).
import { select, Store } from '@ngrx/store';
this.searchQuery$ = store.select(fromBooks.selectSearchQuery);
```

I would've expected that the unused import also had a fixer, but it seems like it does not have one - thus after this fix we end up with a new warning 😅

rafaelss95:
Hmm, this is tricky as there is no way to actually know if there are other occurrences of `select`. Coincidentally, I talked about this with James here https://github.com/angular-eslint/angular-eslint/pull/606#discussion_r697892705. In this specific case, which had the intention to remove the imports from lifecycle interfaces, he suggested that we do a string check and then remove if there was only one occurrence (the import) and so we did... however, for this case here I don't know if this approach would be viable, since a search for 'select' would still return several occurrences, as it only changed the position/way it is called and we can't really trust on something like `.select` or `(select)` because of formatting 😞.

PR #191, opened 2021-09-07
feat(no-effect-decorator-and-creator): add fixer/suggest

rafaelss95:
Part of #179.


timdeschryver (approved):

timdeschryver:
Thanks for all the PRs @rafaelss95 

I also just noticed that you're using your fork of this repo.
If you want, you should also be able to create branches directly from this repo, imho this makes it a bit easier to keep your branches up to date. Don't feel obligated to do it, if you prefer to use your fork.

tests/rules/no-effect-decorator-and-creator.test.ts
```diff
@@ -1,40 +1,77 @@
...
+        }`,
+      errors: [
+        {
+          column: 1,
+          endColumn: 5,
+          line: 4,
+          messageId: noEffectDecoratorAndCreator,
+          suggestions: [
+            {
+              messageId: noEffectDecoratorAndCreatorSuggest as MessageIds,
```
rafaelss95, line 63: Forgot to comment about this... it's interesting that we need this typecast here. If we remove the compiler isn't satisfied anymore...

Error without typecast

It seems like the `RuleTester.run` isn't being able to catch all `MessageIds` properly if it's a union type 🤔 

bug in RuleTester.run?

PR #193, opened 2021-09-07
 fix: remove unused imports after fix/suggest

rafaelss95:
Here's an attempt to remove the unused imports. PTAL, @timdeschryver :)

### DIFF
--- a/src/rules/effects/no-effect-decorator-and-creator.ts
+++ b/src/rules/effects/no-effect-decorator-and-creator.ts
@@ -1,4 +1,4 @@
-import type { TSESTree } from '@typescript-eslint/experimental-utils'
+import type { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils'
 import { ESLintUtils } from '@typescript-eslint/experimental-utils'
 import path from 'path'
 import {
@@ -7,6 +7,9 @@ import {
   effectDecorator,
   getDecorator,
   getDecoratorArguments,
+  getImportDeclarations,
+  getImportRemoveFix,
+  MODULE_PATHS,
 } from '../../utils'
 
 export const noEffectDecoratorAndCreator = 'noEffectDecoratorAndCreator'
@@ -39,6 +42,8 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   },
   defaultOptions: [],
   create: (context) => {
+    const sourceCode = context.getSourceCode()
+
     return {
       [`${effectCreator}:has(${effectDecorator})`](
         node: TSESTree.ClassProperty,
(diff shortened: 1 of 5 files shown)


timdeschryver (approved):
Let's ship this!
Thanks again @rafaelss95 !

timdeschryver:
I've tried this again in a project and it seems like it works if you just have 1 usage.
When there are multiple usages of `select`, the import doesn't get removed.
My guess is that this is because the ESLint builds up a queue of fixes instead of running them directly.
For example, for 4 usages, the fixer gets invoked 4 times and the `totalPipeSelectOccurrences` count remains 4 for every fix.

timdeschryver:
I fiddled with this and I think the only way to fix this is a bit dirty, but it seems to work... 
The trick is to keep track of all violations and of all fixes, later we can verify if this count is the same in a `Program:exit`, and then add a fix to remove the import. Something like this, thoughts?
The current implementation is just to verify if this would work, and needs to be polished if we want to proceed with this

```ts
let reportCounter = 0
let fixCounter = 0
let importFixer: () => any[] = () => []
let importNode: TSESTree.ImportSpecifier | null = null

return {
        [pipeableSelect(storeName)](node: CallExpression) {
          reportCounter++

          context.report({
            node: node.callee,
            messageId: methodSelectMessageId,
            fix: (fixer) => {
              fixCounter++

              importFixer = () => {
                const importDeclarations =
                  getImportDeclarations(node, MODULE_PATHS.store) ?? []

                return getImportRemoveFix(
                  sourceCode,
                  importDeclarations,
                  'select',
                  fixer,
                )
              }

              return getOperatorToMethodFixes(node, sourceCode, fixer)
            },
          })
        },
        ['ImportSpecifier[imported.name="select"]'](
          node: TSESTree.ImportSpecifier,
        ) {
          importNode = node
        },
        ['Program:exit']() {
          if (
            importNode &&
            importFixter &&
            reportCounter &&
            reportCounter == fixCounter
          ) {
            context.report({
              node: importNode,
              messageId: methodSelectMessageId,
              fix: importFixer,
            })
          }
        }
```

rafaelss95:
> I've tried this again in a project and it seems like it works if you just have 1 usage.
> When there are multiple usages of `select`, the import doesn't get removed.
> My guess is that this is because the ESLint builds up a queue of fixes instead of running them directly.
> For example, for 4 usages, the fixer gets invoked 4 times and the `totalPipeSelectOccurrences` count remains 4 for every fix.

Sad, I imagined this could happen 😢.

---

I liked your implementation, seems really promising. I'm sorry to play devil's advocate, but don't you think we're worrying too much about this import removal? Reading the [offical docs for ESLint fixes](https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes), we found this:

> Best practices for fixes:

> - Avoid any fixes that could change the runtime behavior of code and cause it to stop working.
> - Make fixes as small as possible. Fixes that are unnecessarily large could conflict with other fixes, and prevent them from being applied.
> - Only make one fix per message. This is enforced because you must return the result of the fixer operation from fix().
> - Since all rules are run again after the initial round of fixes is applied, it's not necessary for a rule to check whether the code style of a fix will cause errors to be reported by another rule.

Aren't we violating the `#2` best practice? As I was writing this, I replied to myself: no, we aren't On the other hand, if we keep `import`, then we would possibly violate best practice `#1` as if someone has enabled `noUnusedLocals` in `tsconfig.json` the code will stop working.

---

In a nutshell, yes: let's do this. Even though the rule gets slower, as we'll go through two more nodes (even before the fixer is called) I think it's the best path we have.

timdeschryver:
@rafaelss95 yea... I also keep swinging between "yes let's add this" and "no, it's too much of a hassle" 😅
I just don't like the warning that it gives after the fix is applied.
I think the noUnusedLocals  rule is a popular one, so removing the unused import would be nice.

I'm going to give this a few days to think on another solution.

rafaelss95:
Another place this would be useful would be the `prefer-concat-latest-from` rule.

Issue #192, opened 2021-09-07
ng-add schematic adds this plugin multiple times


timdeschryver:
## Behavior

I've run the `ng add` schematic on the NgRx repo, and the result was that the NgRx ESLint plugin was added multiple times to the same config.

## Code snippet

```json
{
      "plugins": [
        "ngrx",
        "ngrx",
        "ngrx"
      ],
      "extends": [
        "plugin:ngrx/recommended",
        "plugin:ngrx/recommended",
        "plugin:ngrx/recommended"
      ]
}
```

and

```json
{
      "plugins": [
        "eslint-plugin-import",
        "@typescript-eslint",
        "ngrx",
        "ngrx",
        "ngrx"
      ],
      "extends": [
        "plugin:ngrx/recommended",
        "plugin:ngrx/recommended",
        "plugin:ngrx/recommended"
      ]
}
```

## I would be willing to submit a PR to fix this issue :heart:

[x] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

PR #194, opened 2021-09-07
feat(prefix-selectors-with-select): add support for `createSelectorFactory`

rafaelss95:
Here's the follow-up PR to handle `createSelectorFactory` as discussed in https://github.com/timdeschryver/eslint-plugin-ngrx/pull/169#discussion_r693381896.

LMK if I'm missing something in the implementation.

### DIFF
--- a/docs/rules/prefix-selectors-with-select.md
+++ b/docs/rules/prefix-selectors-with-select.md
@@ -39,7 +39,9 @@ Examples of **correct** code for this rule:
 export const selectFeature = createSelector((state: AppState) => state.feature)
 export const selectFeature: MemoizedSelector<any, any> = (state: AppState) =>
   state.feature
+
 const selectFeature = createFeatureSelector<FeatureState>(featureKey)
+
 export const selectThing = (id: string) =>
   createSelector(selectThings, (things) => things[id])
 ```
--- a/docs/rules/prefix-selectors-with-select.md
+++ b/docs/rules/prefix-selectors-with-select.md
@@ -37,6 +37,7 @@ Examples of **correct** code for this rule:
 
 ```ts
 export const selectFeature = createSelector((state: AppState) => state.feature)
+
 export const selectFeature: MemoizedSelector<any, any> = (state: AppState) =>
   state.feature
 
--- a/docs/rules/prefix-selectors-with-select.md
+++ b/docs/rules/prefix-selectors-with-select.md
@@ -13,7 +13,8 @@ Examples of **incorrect** code for this rule:
 export const feature = createSelector((state: AppState) => state.feature)
 
 // ⚠ Usage of a selector without any description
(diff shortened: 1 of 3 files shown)


docs/rules/prefix-selectors-with-select.md
```diff
@@ -30,6 +40,8 @@ export const selectFeature = createSelector((state: AppState) => state.feature)
 export const selectFeature: MemoizedSelector<any, any> = (state: AppState) =>
   state.feature
 const selectFeature = createFeatureSelector<FeatureState>(featureKey)
+export const selectThing = (id: string) =>
```
timdeschryver, line 43: Nit, to keep things inline we should add linebreaks between the correct examples 😅


```suggestion

export const selectThing = (id: string) =>
```

docs/rules/prefix-selectors-with-select.md
```diff
@@ -30,6 +40,8 @@ export const selectFeature = createSelector((state: AppState) => state.feature)
 export const selectFeature: MemoizedSelector<any, any> = (state: AppState) =>
   state.feature
 const selectFeature = createFeatureSelector<FeatureState>(featureKey)
```
timdeschryver, line 42:
```suggestion

const selectFeature = createFeatureSelector<FeatureState>(featureKey)
```

docs/rules/prefix-selectors-with-select.md
```diff
@@ -29,7 +39,11 @@ Examples of **correct** code for this rule:
 export const selectFeature = createSelector((state: AppState) => state.feature)
 export const selectFeature: MemoizedSelector<any, any> = (state: AppState) =>
```
timdeschryver, line 40:
```suggestion

export const selectFeature: MemoizedSelector<any, any> = (state: AppState) =>
```

timdeschryver (approved):
LGTM,
I think we can also add a suggestion for this rule in #192 , WDYT?

rafaelss95:
> LGTM,
> I think we can also add a suggestion for this rule in #192 , WDYT?

Isn't this issue about calling schematics multiple times? I think I didn't understand the relationship with this specific rule. Could you clarify it? 😅

timdeschryver:
Oh yea, that's not clear at all and it also seems that the wrong issue number is linked - sorry about that.
What I mean is that we could create a suggestion to rename the wrong named selector to the correct, e.g. `getCount` to `selectCount`. It should've been linked to #179

rafaelss95:
Aha, now I got it 😄 What should we do if I have for example these cases:

- `const select = createSelector(...)` => `const selectSelect = createSelector(...)`?
- `const loaderselector = createSelector(...)` => - `const selectLoaderselector = createSelector(...)`?

timdeschryver:
That seems correct to me @rafaelss95

PR #197, opened 2021-09-13
feat(use-consistent-global-store-name): add suggestion

rafaelss95:


tests/rules/use-consistent-global-store-name.test.ts
```diff
@@ -22,35 +23,75 @@ ruleTester().run(path.parse(__filename).name, rule, {
...
+          column: 30,
+          endColumn: 44,
+          line: 4,
+          messageId: useConsistentGlobalStoreName,
+          data: {
+            storeName: 'store',
+          },
+          suggestions: [
+            {
+              messageId: useConsistentGlobalStoreNameSuggest as MessageIds,
```
rafaelss95, line 52: Still don't get why TS isn't able to infer this 😅

timdeschryver: I have no clue; I've searched a solution for it... but yea... 🤯

timdeschryver (approved):

PR #199, opened 2021-09-14
feat(on-function-explicit-return-type): add suggestion

rafaelss95:
Part of #179.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   },
   "dependencies": {
     "@angular-devkit/schematics": "^12.2.2",
-    "@typescript-eslint/experimental-utils": "^4.29.2",
+    "@typescript-eslint/experimental-utils": "^4.31.0",
     "eslint-etc": "^4.1.0",
     "eslint-plugin-rxjs": "^3.3.6",
     "tsutils-etc": "^1.3.4"
@@ -66,8 +66,8 @@
     "@types/eslint": "^7.28.0",
     "@types/node": "^14.14.7",
     "@types/prettier": "^2.3.2",
-    "@typescript-eslint/eslint-plugin": "^4.31.0",
-    "@typescript-eslint/parser": "^4.31.0",
+    "@typescript-eslint/eslint-plugin": "^4.31.1",
+    "@typescript-eslint/parser": "^4.31.1",
     "common-tags": "^1.8.0",
     "cpy-cli": "^3.1.1",
     "eslint": "^7.32.0",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   },
   "dependencies": {
     "@angular-devkit/schematics": "^12.2.2",
-    "@typescript-eslint/experimental-utils": "^4.31.0",
+    "@typescript-eslint/experimental-utils": "^4.31.1",
     "eslint-etc": "^4.1.0",
     "eslint-plugin-rxjs": "^3.3.6",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -50,7 +50,7 @@
   },
   "dependencies": {
     "@angular-devkit/schematics": "^12.2.2",
-    "@typescript-eslint/experimental-utils": "^4.29.2",
+    "@typescript-eslint/experimental-utils": "^4.31.0",
```
rafaelss95, line 53: I forgot to update this package along with others before 😞

src/rules/store/on-function-explicit-return-type.ts
```diff
@@ -16,22 +20,54 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+  sourceCode: Readonly<TSESLint.SourceCode>,
+  fixer: TSESLint.RuleFixer,
+): TSESLint.RuleFix | readonly TSESLint.RuleFix[] {
+  const [firstParam] = params
+  const [lastParam] = params.slice(-1)
+  const previousToken = sourceCode.getTokenBefore(firstParam)
+  const isParenthesized =
+    previousToken && ASTUtils.isOpeningParenToken(previousToken)
+
+  if (isParenthesized) {
```
rafaelss95, line 64: I tried to use [`ASTUtils.isParenthesized`](https://github.com/typescript-eslint/typescript-eslint/blob/1c1b572c3000d72cfe665b7afbada0ec415e7855/packages/experimental-utils/src/ast-utils/eslint-utils/astUtilities.ts#L113-L116) but it doesn't work for this case, as it only returns `true` if a node has `parenthesis` before **and** after, which isn't necessarily our case, as we can have, for example, `on(action, (a, b) => ...`, so neither param `a`, nor param `b` would receive `true` as a response from this `function`.

src/rules/store/on-function-explicit-return-type.ts
```diff
@@ -16,22 +20,54 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
       description: 'On function should have an explicit return type.',
       recommended: 'warn',
+      suggestion: true,
     },
     schema: [],
     messages: {
-      [messageId]:
+      [onFunctionExplicitReturnType]:
         'On functions should have an explicit return type when using arrow functions: `on(action, (state): State => {}`.',
+      [onFunctionExplicitReturnTypeSuggest]: 'Add explicit return type.',
```
timdeschryver, line 29: I think we should mention that we've chosen `State` as return type, but that the user should update it if needed.
What do you think?

rafaelss95: Cool. What's your suggestion?

timdeschryver: Ha! that's a good question (and of course it's easy for me to point this out without providing something better) 😅
Something like this?

"Add explicit return type (By default `State`, when the interface is named different you need to manually correct the return type)."

rafaelss95: What about this? Feel free to apply the suggestion if you agree :)

```suggestion
      [onFunctionExplicitReturnTypeSuggest]: 'Add explicit return type `State` (if the interface is named differently you need to manually correct the return type).',
```

timdeschryver (approved):

PR #201, opened 2021-09-14
feat(schematics): ask for config while installing the plugin

timdeschryver:


src/schematics/ng-add/index.ts
```diff
@@ -54,14 +55,13 @@ ${detailsContent}
 `)
     }
   }
-}
```
rafaelss95, line 57: Is this line removal inteded? :see_no_evil:

rafaelss95 (approved):

src/schematics/ng-add/index.ts
```diff
@@ -54,14 +55,13 @@ ${detailsContent}
...
-
```
rafaelss95, line 58: Is this line removal inteded? :see_no_evil:

rafaelss95 (approved):

src/schematics/ng-add/index.ts
timdeschryver: It's not a removal. the method is moved into the scope of the default method.

PR #203, opened 2021-09-14
feat(no-effects-in-providers): add fixer

rafaelss95:
Part of #179.

In addition to the fixer, it fixes the same cases we did for other rules in #186 and #190, that is, the report should only happen within the same `@NgModule` metadata.


src/rules/effects/no-effects-in-providers.ts
```diff
@@ -31,26 +32,46 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+  sourceCode: Readonly<TSESLint.SourceCode>,
+  fixer: TSESLint.RuleFixer,
+  node: TSESTree.Identifier,
+) {
+  const nextToken = sourceCode.getTokenAfter(node)
+  const isNextTokenComma = nextToken && ASTUtils.isCommaToken(nextToken)
+  return [
+    fixer.remove(node),
+    ...(isNextTokenComma ? [fixer.remove(nextToken)] : []),
+  ]
```
rafaelss95, line 76: I could have done it this way

```ts
return fixer.removeRange([
  node.range[0],
  isNextTokenComma ? nextToken.range[1] : node.range[1],
])
```

... but then possibly unnecessary changes would be made, such as removing comments indiscriminately.

timdeschryver (commented):
For me these changes are good, but the build things different :(

rafaelss95:
Yes, it was failing because the `@typescript-eslint/experimental-utils` wasn't in the latest. It should be good now after the merge of #199.

timdeschryver (approved):

PR #205, opened 2021-09-15
feat(prefix-selectors-with-select): add suggestion

rafaelss95:
Part of #179.


src/rules/store/prefix-selectors-with-select.ts
```diff
@@ -17,18 +21,22 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
         'The selector should start with "select", for example "selectThing".',
       recommended: 'warn',
+      suggestion: true,
     },
     schema: [],
     messages: {
-      [messageId]: 'The selector should start with "select".',
+      [prefixSelectorsWithSelect]: 'The selector should start with "select".',
+      [prefixSelectorsWithSelectSuggest]:
+        'Prefix the selector with the word "select" => `{{name}}`.',
```
rafaelss95, line 30: Feel free to suggest a better one :)

src/rules/store/prefix-selectors-with-select.ts
```diff
@@ -38,9 +46,52 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+              },
+            },
+          ],
         })
       },
     }
   },
 })
+
+function getSuggestedName(name: string) {
```
rafaelss95, line 69: So... this is a proposal... this function basically tries to deduce what the best name would be. Not sure if it's a good idea or if we should just append `select` + capitalize the first existent letter and that's fine or even better... if you have something in mind.

In any case, I'm all ears.

tests/rules/prefix-selectors-with-select.test.ts
```diff
@@ -24,48 +25,175 @@ ruleTester().run(path.parse(__filename).name, rule, {
...
-                     ~~~~~~~~~~~~                     [${messageId}]
-          defaultMemoize(
-            projectionFun,
-            orderDoesNotMatterComparer,
-            orderDoesNotMatterComparer,
-          ),
-        )`,
-    ),
+    {
+      code: stripIndent`
```
rafaelss95, line 29: The bad with suggestions is that the tests become really verbose with this big `errors` array 😞

tests/rules/prefix-selectors-with-select.test.ts
```diff
@@ -24,48 +24,175 @@ ruleTester().run(path.parse(__filename).name, rule, {
...
+          column: 14,
+          endColumn: 22,
+          line: 1,
+          messageId: prefixSelectorsWithSelect,
+          data: {
+            name: 'selectCount',
+          },
+          suggestions: [
+            {
+              messageId: prefixSelectorsWithSelectSuggest,
```
rafaelss95, line 41: For whatever reason, I noticed that the typecast (`as MessageIds`) is only necessary when you have both `fromFixture` and object-like tests within the `invalid` section 🤷‍♂️

package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@angular-devkit/schematics": "^12.2.2",
     "@typescript-eslint/experimental-utils": "^4.31.1",
-    "eslint-etc": "^4.1.0",
+    "eslint-etc": "^4.2.4",
```
rafaelss95, line 55: Even without this updating, when I ran `npm i`, all the changes in `package-lock.json` made on #212, are reverted as you can see above. Thoughts?

src/rules/store/prefix-selectors-with-select.ts
timdeschryver: Nit, I think "word" can be left out

```suggestion
        'Prefix the selector with the "select" => `{{name}}`.',
```

src/rules/store/prefix-selectors-with-select.ts
timdeschryver: I think that's a great idea! 
I say, keep it!

src/rules/store/prefix-selectors-with-select.ts
```diff
@@ -38,9 +46,52 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+    },
+  )
+
+  if (name !== possibleReplacedName) {
+    return possibleReplacedName
+  }
+
+  // Ex: 'getCount' => 'selectCount'
+  possibleReplacedName = name.replace(
+    /[a-z\d]+([A-Z][A-Za-z\d]+)/,
```
timdeschryver, line 85: If I'm not mistaken, this would also replace other words as `get`, right?
e.g. `customersWithOrders` would become `selectWithOrders` .
If that's the case, I don't think it's a good idea and that we should just check for `get`

rafaelss95: Hmm yes, I intentionally did this. But after thinking more I'll go with your suggestion.

timdeschryver (approved):

timdeschryver:
Yikes! Conflict! ⚠

PR #209, opened 2021-09-15
feat(avoid-duplicate-actions-in-reducer): add suggestion

rafaelss95:
Part of #179.


src/utils/helper-functions/utils.ts
```diff
@@ -188,6 +189,19 @@ export function getImportRemoveFix(
   ])
 }
 
+export function getNodeToCommaRemoveFix(
```
rafaelss95, line 192: Not sure if it's a good name 😄

timdeschryver: what about removeNodeAndTrailingComma ?

rafaelss95: I think we can move on with this name, as it would be probably replaced by `ASTUtils. getNodeToCommaRemoveFix` (yes, for lack of a better name, I used that exact name there 😆) from `@angular-eslint/utils` in a short period. WDYT?

timdeschryver: 👍👍

timdeschryver (approved):

PR #210, opened 2021-09-15
build: enable some rules

rafaelss95:
This enables the rules:

- [**`@typescript-eslint/no-unnecessary-type-assertion`**](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unnecessary-type-assertion.md);
- [**`@typescript-eslint/prefer-reduce-type-parameter`**](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-reduce-type-parameter.md)

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "build": "tsc",
     "postbuild": "npm run copy:schematics",
     "copy:schematics": "cpy ./schematics/**/*.json ../dist --cwd=src --parents",
-    "lint:all": "eslint ./src/**/*.ts ./tests/**/*.ts --fix",
+    "lint:all": "eslint ./scripts/**/*.ts ./src/**/*.ts ./tests/**/*.ts --fix",
     "prettier:all": "prettier --write .",
     "g:config": "ts-node ./scripts/generate-config.ts",
     "g:readme": "ts-node ./scripts/generate-readme.ts",
--- a/scripts/generate-config.ts
+++ b/scripts/generate-config.ts
@@ -13,12 +13,15 @@ const getRules = (
   predicate: (rule: RuleModule) => boolean,
   setting = (rule: RuleModule) => rule.meta.docs?.recommended || 'warn',
 ) =>
-  Object.entries(rules).reduce((rules, [ruleName, rule]) => {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "build": "tsc",
     "postbuild": "npm run copy:schematics",
     "copy:schematics": "cpy ./schematics/**/*.json ../dist --cwd=src --parents",
-    "lint:all": "eslint ./src/**/*.ts ./tests/**/*.ts --fix",
+    "lint:all": "eslint ./scripts/**/*.ts ./src/**/*.ts ./tests/**/*.ts --fix",
```
rafaelss95, line 43: I've also enabled this, LMK if you think it's a good ideia...

scripts/generate-config.ts
```diff
@@ -13,12 +13,15 @@ const getRules = (
   predicate: (rule: RuleModule) => boolean,
   setting = (rule: RuleModule) => rule.meta.docs?.recommended || 'warn',
 ) =>
-  Object.entries(rules).reduce((rules, [ruleName, rule]) => {
-    if (predicate(rule)) {
-      rules[`${RULE_NAME_PREFIX}${ruleName}`] = setting(rule)
-    }
-    return rules
-  }, {} as Record<string, string>)
+  Object.entries(rules).reduce<Record<string, string>>(
```
timdeschryver, line 16: I'm not a big fan of using generics to be honest 😅
I prefer casting it myself, because generics can change and have breaking changes.
It probably won't be the cause with `.reducer` though.

package.json
timdeschryver: Yea, that's fine 👍

timdeschryver (approved):

scripts/generate-config.ts
rafaelss95: Oh, I prefer the opposite, but in that case I can revert these changes if you want to. 😄

timdeschryver: Nah, it's fine.

PR #211, opened 2021-09-15
feat(no-dispatch-in-effects): add suggestion

rafaelss95:
Part of #179.


src/utils/selectors/index.ts
```diff
@@ -46,7 +44,12 @@ export const effectsInNgModuleProviders =
   `${ngModuleProviders} Identifier` as const
 
 export const actionDispatch = (storeName: string) =>
-  `ExpressionStatement > CallExpression:matches([callee.object.name='${storeName}'][callee.property.name='dispatch'], [callee.object.object.type='ThisExpression'][callee.object.property.name='${storeName}'][callee.property.name='dispatch'])` as const
+  `CallExpression:matches([callee.object.name='${storeName}'][callee.property.name='dispatch'], [callee.object.object.type='ThisExpression'][callee.object.property.name='${storeName}'][callee.property.name='dispatch'])` as const
+
+export const dispatchInEffects = (storeName: string) =>
+  `${createEffectExpression} ${actionDispatch(
+    storeName,
+  )} > MemberExpression` as const
```
rafaelss95, line 52: Instead of repeating the selectors, here I'm reusing the `createEffectExpression` and `actionDispatch`.

timdeschryver (commented):
How would this suggestion work with multiple dispatches?
I think it would be good to have an example in the tests

```ts
effectNOK = createEffect(() => this.actions.pipe(
  ofType('PING'),
  map(() => {
    this.store.dispatch({ type: 'PONG' });
    this.store.dispatch({ type: 'PONG2' });
    this.store.dispatch({ type: 'PONG3' });
    return {type: 'PONG4'};
  })
))
```

I'm asking because I think it would be easier (and perhaps better?) to just remove the whole expression.

So it would become


```ts
effectNOK = createEffect(() => this.actions.pipe(
  ofType('PING'),
  map(() => {
    return {type: 'PONG4'};
  })
))
```

For the tests that are already covered it would become:


```ts
effectNOK = createEffect(() => this.actions.pipe(
  ofType('PING'),
  map(() => {
   // because nothing is returned `createEffect` will complain (if dispatch isn't set to false)
    return;
  })
))
```

Thoughts?

rafaelss95:
Good question! As we consider each `dispatch` individually, the suggestion doesn't care about that, that is, it would be the same for one or several cases, and therefore:

```ts
map(() => {
  ;
  ;
  ;
  return {type: 'PONG4'};
})
```

Also note that the suggestion only keeps the action part if `store.dispatch` if it's **preeceded** by `ArrowFunctionExpression` or `ReturnStatement`.

The interesting thing we doesn't have a case with semicolon 😆. Can we can improve it by removing the semicolon as well? Yes, but I don't know if we really should, as it doesn't generate any compiler error and formatters like `prettier` can perform this job better.

From https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions:

> Note: Suggestions will be applied as a stand-alone change, without triggering multipass fixes. Each suggestion should focus on a singular change in the code and should not try to conform to user defined styles. For example, if a suggestion is adding a new statement into the codebase, **it should not try to match correct indentation, or conform to user preferences on presence/absence of semicolons**. All of those things can be corrected by multipass autofix when the user triggers it.

---

About adding the test itself with multiple cases, I was avoiding this as much as possible in a single test, as it significantly increases the size of the test becoming easy to get lost with so many lines. We could however create a similar test, but with one `dispatch`, like this:

```ts
effectNOK = createEffect(() =this.actions.pipe(
  ofType('PING'),
  map(() ={
    this.store.dispatch({ type: 'PONG' });
    return {type: 'PONG4'};
  })
))
```

LMK if it answers your question and WDYT about the suggestion.

timdeschryver:
I would love to see the test you proposed in our test cases.

> The interesting thing is that we doesn't have a case with semicolon

Yea, that will be removed by another rule/prettier, that's not something we should worry about 🙃

rafaelss95:
Well, I just noticed that we already have this case. See [here](https://github.com/timdeschryver/eslint-plugin-ngrx/pull/211/files#diff-7ca9e826b61372dffbd27899143161b21db7242a240888e6a5dfe063773cb016R120).

tests/rules/no-dispatch-in-effects.test.ts
```diff
@@ -26,27 +45,155 @@ ruleTester().run(path.parse(__filename).name, rule, {
...
+    ),
+    fromFixture(
+      stripIndent`
+      import { Store } from '@ngrx/store'
+      @Injectable()
+      export class FixtureEffects {
+        effectNOK3 = createEffect(
+          () =>
+            this.actions.pipe(
+              ofType('PING'),
```
rafaelss95, line 160: Btw, wouldn't it be good if we create a rule to avoid this kind of use?

timdeschryver: That's good for me 👍
If we do that, I think adding a similar one for `.dispatch()` would also make sense

timdeschryver (approved):

rafaelss95:
@timdeschryver I guess you missed my last comment [here](https://github.com/timdeschryver/eslint-plugin-ngrx/pull/211#issuecomment-922109311), right? 😅

timdeschryver:
@rafaelss95 I suppose you're right. By looking at the gist it also doesn't seem too complex to do (I thought it was).

I sometimes look at stuff on my phone, and then the notification disappears... I always think that I can remembers questions but if I miss something feel free to call me out.

tests/rules/no-dispatch-in-effects.test.ts
rafaelss95: Actually `store.dispatch` doesn't accept a `string`, only `Action`.

timdeschryver: Correct, but it does support `store,dispatch({type: 'an action'})`.
And maybe we should also lint the old syntax (`store,dispatch(new Action())`) but this might be too harsh, so we could move  it behind a config?

Issue #206, opened 2021-09-15
Setup an internal package for repo maintenance


rafaelss95:
To track the plans about the linter for repo maintenance discussed a bit at https://github.com/timdeschryver/eslint-plugin-ngrx/pull/185#issuecomment-913007070.

timdeschryver:
Feel free to take this one @rafaelss95

rafaelss95:
Do you have something in mind? Like.. where should I start?

PR #212, opened 2021-09-16
style: add internal rules

timdeschryver:
Starter of #206

Some decisions I made:

- create a new folder `internal-rules` to put the rules in
- create the rules in TS and build them to JS, otherwise the internal rules would be written differently in comparison to our public rules. (sadly, we need to compile to JS, simply referencing TS rules doesn't work AFAIK)
- commit the JS files for the internal rules, otherwise we would've to build it every time + this might be harder when a new contributor comes in 
- we could automate the process and add a pre-commit + ci check to see if the internal rules are up to date, this way we're sure everyone works on the latest version


rafaelss95 (approved):
Awesome! ❤️ 🚀

rafaelss95:
I modified the `first-rule.ts` to grab all `CallExpression` as a test to see how it's working and for some reason it isn't linting the files correctly. Here's the result from `eslint ./src/**/*.ts --fix` (and yes I bundled it using the `npm run build:internal-rules`):

Captura de Tela 2021-09-18 às 00 31 00
s

timdeschryver:
Hmmm that's strange, I think I also tested this and it linted the files correctly.
I'll have a look later 🤔

timdeschryver:
Are you sure? 
It seems to be working here 😅

image

rafaelss95:
See #219.

PR #198, opened 2021-09-13
feat(no-multiple-global-stores): add suggestion

rafaelss95:

PR #213, opened 2021-09-18
refactor(prefer-one-generic-in-create-for-feature-selector): add missing suggestion for all nodes

rafaelss95:
Currently, the suggestion only applies to `TSTypeReference`, but I don't see any reason to have this restriction. A real example of another type of `node` can be found here:

https://github.com/ngrx/platform/blob/0bb49409413b784b1cc55c2b59e67ed760041aa5/projects/ngrx.io/content/examples/testing-store/src/app/state/books.selectors.ts#L10-L13

Taking the fact I'm writing about it... wouldn't be a good ideia to add this plugin to the `ngrx`'s eslint file in that repo?


package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@angular-devkit/schematics": "^12.2.2",
     "@typescript-eslint/experimental-utils": "^4.31.1",
-    "eslint-etc": "^4.1.0",
+    "eslint-etc": "^4.2.4",
```
rafaelss95, line 55: Even without this updating, when I ran `npm i`, all the changes in `package-lock.json` made on #212, are reverted as you can see above. Thoughts?

timdeschryver: Oh yikes, I didn't pay attention to the lock file...
It' because I'm using npm v7, which uses a different package-lock structure.
I'm guessing you're on a lower version, using the "older" version of the package-lock

rafaelss95: Thanks! I've updated to npm v7 and it seems to working very well now. And it also explains the surprise I had [here](https://github.com/timdeschryver/eslint-plugin-ngrx/pull/198#discussion_r707429351)!

now

timdeschryver: 😂

timdeschryver (approved):

rafaelss95:
Do you have any thoughts about this question:

> Taking the fact I'm writing about it... wouldn't be a good ideia to add this plugin to the ngrx's eslint file in that repo?

timdeschryver:
@rafaelss95 Yea, that might be good. Only for the example app though.

PR #214, opened 2021-09-18
test(use-consistent-global-store-name): add missing `data` for suggestions

rafaelss95:


tests/rules/use-consistent-global-store-name.test.ts
```diff
@@ -81,6 +84,9 @@ ruleTester().run(path.parse(__filename).name, rule, {
           suggestions: [
             {
               messageId: useConsistentGlobalStoreNameSuggest,
+              data: {
```
timdeschryver, line 87: Just out of curiosity, how come that you created a PR for this now?
👀 is it because it gave a warning/error now (maybe with the latest versions), or did you spot that we didn't add the suggestion before? Previously, I only provided the `data` in the rule itself and not in suggestions/fixers, I think...

timdeschryver (approved):

tests/rules/use-consistent-global-store-name.test.ts
rafaelss95: Hehe, I was experimenting something with `convertAnnotatedSourceToFailureCase` and by coincidence for **this** file, it complaining saying:

> AssertionError [ERR_ASSERTION]: Hydrated message "Global store should have the name `{{ storeName }}`." does not match "Global store should have the name `store`."

PR #215, opened 2021-09-18
docs: add 'requires type information' info

rafaelss95:
There is currently no way to know in advance whether the rule requires type information unless the user has some knowledge and looks at the source code.


src/rules/effects/no-multiple-actions-in-effects.ts
```diff
@@ -21,6 +21,7 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
       category: 'Best Practices',
       description: 'An `Effect` should not return multiple actions.',
       recommended: 'warn',
+      requiresTypeChecking: true,
```
rafaelss95, line 24: While I understand the point of `avoid-cyclic effects` being a rule that requires type checking, this one I don't understand why it needs to be either. 
I mean, users may have a [performance loss](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/TYPED_LINTING.md#performance) to report a rare case like this:

```ts
export class Effects {
  test$ = createEffect(() =>
    this.actions$.pipe(concatMap(() => {
      const actions: Action[] = [];
      return actions;
    }))
  )
}
```

Unless I'm missing something, IMHO it shouldn't require type-checking as the other rules we already have.

timdeschryver: It's indeed for that case. 
I've added this because I saw the following code in one of our projects:
The only way I could report this was by using type checking.

```ts
export class Effects {
  test$ = createEffect(() =>
    this.actions$.pipe(concatMap(() => {
      const actions: Action[] = [];
      if(condition) actions.push(action1);
      if(condition2) actions.push(action2);
      if(condition3) actions.push(action3);
      return actions;
    }))
  )
}
```

timdeschryver (commented):
Do you think that this information is useful to our users when they're reading the README?
I personally don't see a benefit.

rafaelss95:
I think so, because of the extra configuration needed (if done manually). [The `@typescript-eslint` does have this](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/README.md#supported-rules).

timdeschryver:
I see, that's interesting thanks @rafaelss95 
If you can resolve the conflict, we'll merge this in 👍
Do you think we should also make a separate config for it, I think that performance boost won't be that great because we just have 2 rules that depend on it. On the other hand, it's also not that hard to create an extra config for it.

rafaelss95:
> If you can resolve the conflict, we'll merge this in 👍

Resolved!

> Do you think we should also make a separate config for it, I think that performance boost won't be that great because we just have 2 rules that depend on it. On the other hand, it's also not that hard to create an extra config for it.

At the moment, I don't have a strong opinion about it. I think we can leave it at that until someone raises a question about it (that is, if that happens).

PR #218, opened 2021-09-19
fix(select-style): remove unused imports after fixing violations

rafaelss95:
@timdeschryver  I took the liberty of submitting this PR as a draft in an attempt to finally clean up any traces of unused imports.

In a short, it basically lookup for the node import `import { ..., select } from '@ngrx/platform'` and then it grab the references that points to this node. Note that the same approach could be used for `no-effect-creator-and-decorator`, `no-effect-decorator` and `prefer-concat-latest-from`. Let me know what you think.

### DIFF
--- a/src/rules/store/select-style.ts
+++ b/src/rules/store/select-style.ts
@@ -5,10 +5,11 @@ import {
   docsUrl,
   findNgRxStoreName,
   getImportAddFix,
-  getImportDeclarations,
   getImportRemoveFix,
   getNearestUpperNodeFrom,
+  isCallExpression,
   isClassDeclaration,
+  isMemberExpression,
   MODULE_PATHS,
   pipeableSelect,
   storeSelect,
@@ -16,7 +17,6 @@ import {
 
 export const methodSelectMessageId = 'methodSelect'
 export const operatorSelectMessageId = 'operatorSelect'
-
 export type MessageIds =
   | typeof methodSelectMessageId
   | typeof operatorSelectMessageId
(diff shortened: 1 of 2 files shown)


src/rules/store/select-style.ts
```diff
@@ -141,43 +138,51 @@ function getMethodToOperatorFixes(
...
-
+  const text = sourceCode.getText(parent)
+  const nextToken = sourceCode.getTokenAfter(parent)
+  const storeNode = callee.object
   return [
-    fixer.removeRange(selectOperatorRange),
-    fixer.insertTextAfterRange(storeRange, `.${text}`),
+    fixer.remove(parent),
+    ...(nextToken ? [fixer.remove(nextToken)] : []),
+    fixer.insertTextAfter(storeNode, `.${text}`),
```
rafaelss95, line 186: I've improved the fixers so it's not dependant on a _well-formatted_ code, instead of decreasing -1 or adding +1 to find the expected range (which may cause a lot of breakage - because it all depends on how one format their code), it looks for the closest node through the `souceCode.getTokenBefore/After`.

timdeschryver (commented):
I like this a lot, smart!
I think it's much better than what we currently have and what was proposed before.
The only downside is that if someone fixes the import, they will get errors because `select` is still being used, correct? I'm fine with that though, just wanted to make sure.

rafaelss95:
Hmm.. I got your concern, but IMHO, fixes, unlike suggestions are usually applied automatically, by saving a file if you've configured your IDE for that, or running via the CLI (manual or pre-commit hook or something like that)... so I assume that it's only a problem if you try to fix every occurrence by hand 🤔.

I don't know... should we bother with this? 😅

timdeschryver:
We shouldn't be bothered with this, at least not for now.
This can be shipped for me 🚀

timdeschryver (approved):

timdeschryver:
Fyi, I just tested this in a real application and it works as a charm! Thanks @rafaelss95 🏆

PR #219, opened 2021-09-19
feat(eslint-plugin-internal): [report-message-format] create rule

rafaelss95:


internal-rules/out/report-message-format.test.js
```diff
@@ -0,0 +1,119 @@
+'use strict'
```
rafaelss95, line 1: Do we really need to versioning these *.js files?

internal-rules/report-message-format.test.ts
```diff
@@ -0,0 +1,80 @@
...
+function ruleTester() {
+  return new TSESLint.RuleTester({
+    parser: resolve('./node_modules/@typescript-eslint/parser'),
+    parserOptions: {
+      project: resolve('./tests/tsconfig.json'),
+      ecmaVersion: 2020,
+      sourceType: 'module',
+    },
+  })
+}
```
rafaelss95, line 16: Ideally it would be better to import this from an `utils` package.

internal-rules/report-message-format.ts
```diff
@@ -0,0 +1,47 @@
...
+    messages: {
+      [messageId]: 'Report message does not match the pattern `{{pattern}}`.',
+    },
+  },
+  defaultOptions: ['^[A-Z].+\\.$'],
+  create: (context, [format]) => {
+    const pattern = RegExp(format)
+
+    return {
+      [`CallExpression[callee.callee.object.name='ESLintUtils'][callee.callee.property.name='RuleCreator'] Property[key.name='meta'] Property[key.name='messages'] :matches(Literal[value!=${pattern}], TemplateLiteral[quasis.length=1][quasis.0.value.raw!=${pattern}])`](
```
rafaelss95, line 34: I tried to simplify this at maximum, hoping someone doesn't uses computed/literal keys 😅

rafaelss95:
Note that in the last commit here (https://github.com/timdeschryver/eslint-plugin-ngrx/pull/219/commits/011b1c480c0891b006adc16a676c05f9cb7482db), I removed the period from the `use-consistent-global-store-name` rule and despite having the error with the VSCode extension and in github annotations, ESLint itself doesn't seem to recognize this as an error and so the CI passes all the checks. See below:

Github annotations

ESLint extension

internal-rules/report-message-format.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ESLintUtils } from '@typescript-eslint/experimental-utils'
+import path from 'path'
+
+export const messageId = 'reportMessageFormat'
+type MessageIds = typeof messageId
+
+type Options = [string]
+
+export default ESLintUtils.RuleCreator(() => '')<Options, MessageIds>({
+  name: path.parse(__filename).name,
```
rafaelss95, line 11: I borrowed the rule name from https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/report-message-format.md, which makes perfect sense to them, since their rule also deals with `context.report({...})`... but for us, who only deal with `messages: { messageId }`, it might make more sense to call it ` message-id-format` or simply ` message format`.

internal-rules/out/report-message-format.test.js
timdeschryver: You mean all the js files, or jusst the tests?

timdeschryver:
How do you run eslint from the CLI?
If I use the following, it seems to work for me.

```ts
npx eslint ./src/rules/effects/no-multiple-actions-in-effects.ts
npm run lint:all
```

internal-rules/out/report-message-format.test.js
rafaelss95: I mean all... I know that we need the compiled files to lint the rules, but can't we have a smarter mechanism rather than versioning these?

internal-rules/report-message-format.test.ts
rafaelss95: > Oh, you wrote tests for internal rules 😅 I didn't think of that.

This is the price you pay for being a lazy person (me 😄), I'd rather write them than keep generating the js files and manually checking if everything is ok.

> The same counts for the `utils` package.
> I think that if we want that, that we need to move some thing around.
> Is it OK to leave things as is, and reevaluate this later if it's needed?

Yeah, for sure.

rafaelss95:
> How do you run eslint from the CLI?
> If I use the following, it seems to work for me.
> 
> ```ts
> npx eslint ./src/rules/effects/no-multiple-actions-in-effects.ts
> npm run lint:all
> ```

Not for me 😞 Why does the CI didn't complain either?

rafaelss95:
Maybe we don't even need to maintain this rule here. Check [this](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/176).

internal-rules/report-message-format.test.ts
timdeschryver: Oh, I see... yea that makes sense.
Would it be easier to just write these rules in JS?

timdeschryver:
> Maybe we don't even need to maintain this rule here. Check this.

Oh, that would be great!

internal-rules/report-message-format.test.ts
timdeschryver: If the answer is yes, we can go ahead and write them.
If you prefer to use TS, we can keep the setup as is with this PR and create a new one and remove the JS files and include a smarter mechanism.
I'm fine either way 👍

While I was writing this, we could also add a third option and remove these internal rules.
And simply have a script that runs on commit, and verifies that the description is how we want it.
This would be identical to how we generate the configs/readme, but it would throw if it detects a mall-formatted description. Maybe this would be the simplest way to create and test it, with the best DX.

rafaelss95:
@timdeschryver Just to be clear: my intention with these rules was to add some of the existing rules at https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin + https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts (to format the template literals without requiring us to do it manually in order to maintain it consistent), **but** now that it looks like they (the `eslint-plugin-eslint-plugin` team) will provide the necessary support, I don't know if we still need it.

And about the literal formatting, we could use [**this**](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1478) for example.

Let me know your thoughts on this.

timdeschryver:
@rafaelss95 since our DX isn't great now, I think it would make sense to use the existing rules from plugins that you've listed?
I think you formatted all of our rules, so we can wait a bit till that PR is released,

rafaelss95:
Let's close this and wait then 🥳

PR #220, opened 2021-09-19
feat(no-dispatch-in-effects): add fixer

rafaelss95:
We **can** offer a fix when the `createEffect` is **not** `{ dispatch: false }` **and** the `store.dispatch` comes **right after** an `ArrowFunctionExpression` or `ReturnStatement`. Other cases stay with the suggestions.


src/rules/effects/no-dispatch-in-effects.ts
```diff
@@ -70,3 +89,36 @@ function getNodeToReport(node: MemberExpressionWithinCallExpression) {
...
+  for (const property of callExpressionArgument.properties) {
+    if (
+      !isProperty(property) ||
+      (property.computed && isIdentifier(property.key))
+    ) {
+      return true
+    }
+
+    if (!hasDispatchFalse && isDispatchFalse(property)) {
+      hasDispatchFalse = true
```
timdeschryver, line 111: Could this also be a short-circuit and just return `true`?
That way, we can get rid of the `hasDispatchFalse` variable

rafaelss95: I don't like this either, but I'm afraid we can't.. because we need to loop over **all** the elements since we can't say that it has a dispatch `false` only looking for `dispatch` property... what if we find a non-`Property` (e.g. `SpreadElement`) or a `Identifier`[computed] (`{ [dispatchKey]: false }`)?

timdeschryver: I think we're talking about something different.
What I mean is that if the condition `isDispatchFalse(property)` passes, that we can return `true` and in that way short-circuit the `for` iteration. In other words, once we know it is a `dispatch: false`, we can ignore the rest of the properties.

rafaelss95: Great! I misunderstood your statement at first. I've adjusted one of the tests to make it clear + used `.some` instead of `for-of`.

tests/rules/no-dispatch-in-effects.test.ts
```diff
@@ -164,7 +156,7 @@ ruleTester().run(path.parse(__filename).name, rule, {
                 return somethingElse()
               }),
             ),
-          { dispatch: false, useEffectsErrorHandler: false },
+          { dispatch: true, useEffectsErrorHandler: false, ...options },
```
rafaelss95, line 159: Note that it's correct to `suggest` here, because although it doesn't have `dispatch: fase` **explicitly**, it may have **implicitly** through `options`.

timdeschryver: Thanks! Just for a future reference, I find these comments useful to have in the code.

```suggestion
          // Note that it's correct to suggest, instead of fix, here, because although it doesn't have dispatch: fase explicitly, it may have implicitly through options
          { dispatch: true, useEffectsErrorHandler: false, ...options },
```

rafaelss95: There's a typo. "fase", instead of "false" :sweat_smile:

And in that case, I prefer to put this comment above the "suggestions" key.

timdeschryver: Ha! 💯
I fixed it locally, and should be pushed now. 😅

timdeschryver (approved):

PR #221, opened 2021-09-20
refactor: use `getNgRx*` to grab all `Store` instances

rafaelss95:
So, I've opened this PR to discuss in practice what something we discussed some time ago https://github.com/timdeschryver/eslint-plugin-ngrx/pull/189#issuecomment-914369815...

The proposal here is to normalize the store retrieval. Instead of getting just the `string` that represents the `Store`, we get the `Identifier`, or rather, not just the first one, but all of them (`Identifier[]`), basically `findNgRx*` (simple `string`) would become `getNgrx*Identifiers` (`Identifier[]`).

With this we'll be able to use this method for the remaining rules that still don't use this method to grab the store(s): `no-multiple-global-stores`, `no-typed-global-store` and `use-consistent-global-store-name`.

---

Question/Statements:

Currently, the vast majority of the rules, except the three mentioned above only cares about the first injected `Store`. Should we apply this to `no-typed-global-store` and `use-consistent-global-store-name` too? Or should we change the behavior of `no-typed-global-store` and `use-consistent-global-store-name` to report only the first match?

You can see `use-selector-for-select`'s changes how it would look like if we apply the option 1 => it basically loop through the stores and build the visitors using `reduce`) and you can also see `no-typed-global-store`'s changes how it would look like if we apply the option 2 => it basically gets the first injected instance as we do in the other rules). Let me know your thoughts.

---

Please, note that this is just a draft and not necessarily must be merged.


src/utils/helper-functions/utils.ts
```diff
@@ -338,6 +343,49 @@ function findCorrespondingNameBy(
...
+      return identifiers.concat(
+        parent.parent.parent as IdentifierWithTypeAnnotation,
+      )
+    }
+
+    return identifiers
+  }, [])
+}
+
+export function getNgRxStoreIdentifiers(
```
rafaelss95, line 383: or maybe

```suggestion
export function getNgRxStores(
```

src/rules/store/use-selector-in-select.ts
```diff
@@ -34,28 +34,31 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
-              messageId,
-            })
-          } else {
-            break
+    return stores.reduce(
+      (visitors, store) => ({
+        ...visitors,
+        [`${pipeableSelect(store.name)}, ${storeSelect(store.name)}`](
+          node: TSESTree.CallExpression,
+        ) {
```
rafaelss95, line 44: This would grab any existing `Store`.

src/rules/store/use-consistent-global-store-name.ts
```diff
@@ -37,14 +36,19 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
-      [`${injectedStore}[name!=${storeName}]`]({
-        loc,
-        name,
-        range,
-        typeAnnotation,
-      }: TSESTree.Identifier & {
-        typeAnnotation: TSESTree.TSTypeAnnotation
-      }) {
+      Program() {
+        const [store] = getNgRxStoreIdentifiers(context) ?? []
```
rafaelss95, line 40: Only the first `Store` would be reported.

src/utils/helper-functions/utils.ts
timdeschryver: `getNgRxStoreIdentifiers is fine for me

timdeschryver (commented):
I like the idea.
Having multiple stores is a bad practice (that's why it's also a rule).
For our rules, I think it makes sense to report all the stores (because the multiple stores rule could be turned off).

Did you check the performance gains? or is this just a gut feeling?

If you can resolve the conflicts, I think that we can merge it

timdeschryver (approved):
Feel free to merge this @rafaelss95 if it's ok for you.

timdeschryver:
Hi @rafaelss95 don't worry about it.
You can refactor existing rules in this PR if you want, just make sure that you commit per rule as that would be easier to review for me (but I see you're already doing that 👍).
Ping me when the PR is ready to be reviewed.

timdeschryver (approved):
Looking good! I have nothing else to say  🚀

PR #225, opened 2021-09-25
refactor: organize `context.report` keys

rafaelss95:
Instead of putting `node` + `messageId` in the end, it puts in the begin, which makes it much more readable IMHO.


src/rules/store/prefix-selectors-with-select.ts
```diff
@@ -45,19 +45,18 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
               column: id.typeAnnotation?.range[0] ?? id.range[1],
             },
           },
-          node: id,
```
rafaelss95, line 48: We already have the `loc`, so `node` shouldn't be provided (and I'm curious why it accepts to pass both without complain).

timdeschryver (approved):

PR #226, opened 2021-09-25
refactor: remove `tsutils-etc` explicit dependency

rafaelss95:


src/utils/helper-functions/typecheck.ts
```diff
@@ -1,50 +0,0 @@
-import type { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils'
```
rafaelss95, line 1: This file was basically a copy from https://github.com/cartant/eslint-etc/blob/02584e770982afca432b9feb2c457483589dcde5/source/get-type-services.ts... so instead of coyping it here, why not use it from `eslint-etc` and avoid the dependency on `tsutils-etc` explicitly (yes, it still there, because `eslint-etc` depends on it, but...)

timdeschryver (approved):

src/utils/helper-functions/typecheck.ts
timdeschryver: I think I might have copied it when it wasn't a part of the public api 🤔
Either way, good catch.

PR #227, opened 2021-09-25
test(no-reducer-in-key-names): ensure the correct behavior for hyphenated keys

rafaelss95:
Instead of using our own implementation, it uses `getPropertyName` from `@typescript-eslint`.

### DIFF
--- a/src/rules/store/no-reducer-in-key-names.ts
+++ b/src/rules/store/no-reducer-in-key-names.ts
@@ -49,6 +49,11 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
               messageId: noReducerInKeyNamesSuggest,
               fix: (fixer) => {
                 const keyName = getRawText(key)
+
+                if (!keyName) {
+                  return null
+                }
+
                 return fixer.replaceText(
                   key,
                   keyName.replace(new RegExp(reducerKeyword, 'i'), ''),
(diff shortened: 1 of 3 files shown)


src/rules/store/no-reducer-in-key-names.ts
```diff
@@ -61,3 +68,19 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+
+function getRange(
+  sourceCode: Readonly<TSESLint.SourceCode>,
+  { computed, key }: TSESTree.Property,
+): TSESTree.Range {
+  if (!computed) {
+    return key.range
+  }
+
+  // In case of ['key'] or [`key`], we remove the brackets.
```
timdeschryver, line 80: Shouldn't we try to keep the style?
Regardless of the style, it could be there for a reason, for example `{['foo-reducer']: : reducer}` is valid while '{ foo-reducer: reducer }' is not.

rafaelss95: Oh, I hadn't thought about that case! It makes perfect sense, I think it's better if we keep our implementation and left it as is then and close this PR.

timdeschryver: +1 for the test.
Returning `null` is also fine for me. 

I also have to take a look at the `ASTUtils` but haven't had the time, perhaps we can use something from it instead of implementing it ourselves.

timdeschryver (approved):

PR #228, opened 2021-09-25
docs: remove `fixable` from rules that are not fixable

rafaelss95:


src/rules/effects/prefer-concat-latest-from.ts
```diff
@@ -25,7 +25,6 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
       recommended: 'warn',
       suggestion: true,
     },
-    fixable: 'code',
```
rafaelss95, line 28: We should add this soon as we've upgrade this rule to provide a `fix`.

src/rules/store/no-typed-global-store.ts
```diff
@@ -3,9 +3,9 @@ import { ESLintUtils } from '@typescript-eslint/experimental-utils'
 import path from 'path'
 import { docsUrl, typedStore } from '../../utils'
 
-export const messageId = 'noTypedStore'
+export const noTypedStore = 'noTypedStore'
```
rafaelss95, line 6: I took the opportunity to change this to keep it the same as we do in the other rules

timdeschryver (approved):

src/rules/effects/prefer-concat-latest-from.ts
timdeschryver: What do you mean?

PR #234, opened 2021-09-29
feat(prefer-concat-latest-from): add fixer

### LINKED ISSUE #179: Add more fixers/suggestions

### LINKED ISSUE #242: Some bugs in rules that deals with `ClassProperty`
## Behavior
As mentioned in https://github.com/timdeschryver/eslint-plugin-ngrx/pull/234#discussion_r718911787, we should investigate/correct (if necessary) those rules.

| Rule                                        | PR <a href="">#<a> |
|---------------------------------------------|--------------------|
| `avoid-cyclic-effects`                      | #246                   |
| `no-dispatch-in-effects`                    | #246                   |
| `no-multiple-actions-in-effects`            | #246                   |
| `prefer-concat-latest-from`                 |                    |
| `prefer-effect-callback-in-block-statement` | #243               |
| `updater-explicit-return-type`              | #244                   |

[ ] No

rafaelss95:
Closes #179, closes #242.

Summary of changes:

- [x] fixes the incorrect suggestion (mentioned at https://github.com/timdeschryver/eslint-plugin-ngrx/issues/179#issuecomment-922388284);
- [x] provides a `fix` instead of `suggest`;
- [x] add tests to cover for non-`ClassProperties` + parametrized effects;
- [x] add the option `strict` to always enforce the replacement (mentioned at https://github.com/timdeschryver/eslint-plugin-ngrx/issues/179#issuecomment-922496217)


docs/rules/prefer-concat-latest-from.md
```diff
@@ -38,6 +38,27 @@ class Effect {
 }
 ```
 
+## Rule Config
```
rafaelss95, line 41: Feel free to enhance this docs.

src/rules/effects/prefer-concat-latest-from.ts
```diff
@@ -8,11 +9,18 @@ import {
...
-export type MessageIds = typeof messageId | typeof messageIdSuggest
+export const preferConcatLatestFrom = 'preferConcatLatestFrom'
+export const preferConcatLatestFromSuggest = 'preferConcatLatestFromSuggest'
+export type MessageIds =
+  | typeof preferConcatLatestFrom
+  | typeof preferConcatLatestFromSuggest
 
-type Options = []
+type Options = [{ readonly strict: boolean }]
+const defaultOptions: Options[number] = { strict: true }
```
rafaelss95, line 19: Should the default be `true` or `false`?

tests/rules/prefer-concat-latest-from.test.ts
```diff
@@ -34,87 +35,122 @@ ruleTester().run(path.parse(__filename).name, rule, {
...
+    fromFixture(
+      stripIndent`
+      import { withLatestFrom } from 'rxjs'
+      class EffectNotOk2 {
+        detail$ = createEffect(() => {
+          return this.actions.pipe(
+            ofType(ProductDetailPage.loaded),
+            concatMap((action) =>
+              of(action).pipe(
+                withLatestFrom(this.store.select$(something), (one, other) => somethingElse())
```
rafaelss95, line 123: Do you see any problem in fixing this case directly?

rafaelss95:
It's worth mentioning that the current suggestion doesn't deal with `withLatestFrom` import and so I kept it as is. I guess it would be a little more complicated than with `select-style`, because here if we have the `strict` option disabled, it would be difficult to distinguish, through the parents, whether or not it is "lazy fetch". Or maybe, you could light me up..

docs/rules/prefer-concat-latest-from.md
timdeschryver: Looking good!

src/rules/effects/prefer-concat-latest-from.ts
timdeschryver: I think we should have `false` as default, we can always reevaluate later.
I also want to test this on a few projects 😅

timdeschryver (commented):
Do you think this should be a breaking version?
Because `concatLatestFrom` is fairly new, and this fixer could break projects that use an older version of NgRx 😅

rafaelss95:
The build failing should be fixed by #236.

---

> Do you think this should be a breaking version? Because `concatLatestFrom` is fairly new, and this fixer could break projects that use an older version of NgRx 😅

Hmm .. probably since we don't have any minimum version of `@ngrx` specified.. maybe it would be interesting to add a minimum version of `ngrx` at the `peerDependencies` section. Also, it would be nice to update the minimum of ESLint and TS (not tested, but both minimum versions are a bit outdated and I don't even know if it works for this project). What do you think?

Something that worth mentioning too is that `@typescript-eslint` uses `semver` to control the flow given a TS version. An example can be found here:

https://github.com/typescript-eslint/typescript-eslint/blob/1c1b572c3000d72cfe665b7afbada0ec415e7855/packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts#L50

tests/rules/prefer-concat-latest-from.test.ts
```diff
@@ -1,120 +1,176 @@
...
+        )
+      })
+    }`,
+  ],
+  invalid: [
+    fromFixture(
+      stripIndent`
+      import { withLatestFrom } from 'rxjs'
+      class EffectNotOk1 {
+        effect$ = createEffect(() =>
```
rafaelss95, line 65: Something that I just noticed while messing with this tests is that currently we're looking for `ClassProperty` in this rule, however... what about this case?

```ts
@Injectable()
export class UserEffects {
  readonly load$: CreateEffectMetadata;

  constructor(private readonly actions$: Actions) {
    this.load$ = createEffect(() =>
      this.actions$.pipe(
        ofType(userActions.load),
        withLatestFrom((action) =>
          this.store.select(fromBooks.selectCollectionBookIds),
        ),
        switchMap(([action, bookCollection]) => of({})),
      ),
    );
  }
}
```

Note that this concern isn't related to the fixer and also not related to this rule only, but all the others that rely on `ClassProperty` (except, of course `no-effect-decorator` and `no-effect-decorator-and-creator`).

timdeschryver:
> Something that worth mentioning too is that @typescript-eslint uses semver to control the flow given a TS version. An example can be found here:

That's awesome! I didn't know we could do that.
We could reuse it and enable/disable this rule based on the TS version, I think in most cases peoples keep the versions of Angular and NgRx the same (and thus also TypeScript). If you agree, we can enable this rule starting from TS v4.2.4, otherwise we can just disable it.

tests/rules/prefer-concat-latest-from.test.ts
timdeschryver: Makes sense to me 👍

rafaelss95:
I'd prefer the third option, but I still believe that we should have `@ngrx/store` as a peerDependency and maybe `rxjs` too, because, to provide the correct fix we may need to know if `withLatestFrom` comes from `rxjs` (>= v7) or from `rxjs/operators` (<= v6). Btw, do you know if there's a way to retrieve the `rxjs` version?

---

And a last question: should this be done in this PR? 🤔

timdeschryver:
> to provide the correct fix we may need to know if withLatestFrom comes from rxjs

How so? Aren't we just verifying the operator name?

> And a last question: should this be done in this PR? 

I do think that we do. I'll see what I can do for it tomorrow and just append my commits to this PR.

rafaelss95:
> How so? Aren't we just verifying the operator name?

LOL. What was I thinking? Forget this 🤦 
 
> I do think that we do. I'll see what I can do for it tomorrow and just append my commits to this PR.

Don't you think it's a bit out of scope of this PR? What about delaying this PR (to avoid an automatic release - if that's the case) and create another addressing these versions concerns?

timdeschryver:
@rafaelss95 that's fine by me 👍

rafaelss95:
Aha, I just remembered why I mentioned that we should care about rxjs version... we provide a fix for `map` (not `withLatestFrom` as I mentioned), however it depends on which version... `rxjs` (>= v7) vs `rxjs/operators` (<= v6).

src/rules/effects/prefer-concat-latest-from.ts
```diff
@@ -21,43 +30,83 @@ export default createRule<Options, MessageIds>({
...
-      [messageId]: 'Use `concatLatestFrom` instead of `withLatestFrom`.',
-      [messageIdSuggest]: 'Replace `withLatestFrom` with `concatLatestFrom`.',
+      [messageId]: `Use \`${concatLatestFromKeyword}\` instead of \`${withLatestFromKeyword}\`.`,
     },
   },
-  defaultOptions: [],
-  create: (context) => {
+  defaultOptions: [defaultOptions],
+  create: (context, [{ strict }]) => {
+    const actionsName = findNgRxEffectActionsName(context)
```
rafaelss95, line 55: Do you see any problem in getting the `Actions` to report only in `Effects` that deal with actions? It was the cleanest way I could think of.

src/rules/effects/prefer-concat-latest-from.ts
```diff
@@ -21,43 +30,83 @@ export default createRule<Options, MessageIds>({
...
+    if (!actionsName) return {}
+
+    const sourceCode = context.getSourceCode()
+    const selector = strict
+      ? (`${createEffectExpression} CallExpression > Identifier[name='withLatestFrom']` as const)
+      : (`${createEffectExpression} ${storeExpression(
```
rafaelss95, line 61: Ehh, maybe we should rename this selector in another PR. Actually `storeExpression` represents something named `XYZ` that is injected in `constructor`, not necessarily `store: Store` or `<accessibility> <readonly> store: Store`.

tests/rules/prefer-concat-latest-from.test.ts
```diff
@@ -1,150 +1,290 @@
...
-            })
-          )
+ruleTester({ ngrxModule: NGRX_MODULE_PATHS.effects, version: '12.1.0' }).run(
+  path.parse(__filename).name,
+  rule,
+  {
+    valid: [
+      {
+        code: `
+import { Actions } from '@ngrx/effects'
```
rafaelss95, line 17: This formatting in tests is something that is widely used in `@typescript-eslint` repo and I think it would be a good idea to use it here as well to avoid trying to format it correctly. It's pretty boring when working with `output` and in general trying to format things manually can even be a real waste of time... so this way we don't need to think about it, we simply align to the left. WDYT?

src/rules/effects/prefer-concat-latest-from.ts
```diff
@@ -21,43 +30,96 @@ export default createRule<Options, MessageIds>({
...
+  if (strict) {
+    return `${createEffectExpression} CallExpression > Identifier[name='withLatestFrom']` as const
+  }
+
+  const actionsName = findNgRxEffectActionsName(context)
+  return actionsName
+    ? (`${createEffectExpression} ${storeExpression(
+        actionsName,
+      )} > CallExpression[arguments.length=1] > Identifier[name='${withLatestFromKeyword}']` as const)
+    : null
```
rafaelss95, line 85: 1. Do you see any problem in getting the `Actions` to report only in `Effects` (`strict: false`) that deal with actions? It was the cleanest way I could think of.
2. Ehh, maybe we should rename this selector in another PR. Actually `storeExpression` represents something named `XYZ` that is injected in `constructor`, not necessarily `store: Store` or `<accessibility> <readonly> store: Store`. The same applies to all selectors who derive from this, i.e. `storeExpressionCallable`.

src/rules/effects/prefer-concat-latest-from.ts
```diff
@@ -21,43 +30,96 @@ export default createRule<Options, MessageIds>({
...
+      importName: concatLatestFromKeyword,
+      moduleName: NGRX_MODULE_PATHS.effects,
+      node,
+    }),
+    ...(isUsingDeprecatedProjectorArgument && nextToken
+      ? [
+          getImportAddFix({
+            fixer,
+            importName: 'map',
+            moduleName: 'rxjs',
```
rafaelss95, line 116: This is the "problem" I was mentioning before... should we use `rxjs` or `rxjs/operators`?

tests/rules/prefer-concat-latest-from.test.ts
```diff
@@ -1,150 +1,276 @@
...
-            })
-          )
+ruleTester({ ngrxModule: NGRX_MODULE_PATHS.effects, version: '12.1.0' }).run(
+  path.parse(__filename).name,
+  rule,
+  {
+    valid: [
+      {
+        code: `
+import { of, withLatestFrom } from 'rxjs'
```
rafaelss95, line 17: This formatting "pattern" in tests is widely used in the `@typescript-eslint` repo and is something I would like to try to use here too to avoid messing while trying to format it correctly. It's pretty boring working with `output` and in general trying to format things manually can actually be a real waste of time... so this way we don't have to think about it, we simply left-align. WDYT?

timdeschryver:
> Aha, I just remembered why I mentioned that we should care about rxjs version... we provide a fix for map (not withLatestFrom as I mentioned), however it depends on which version... rxjs (>= v7) vs rxjs/operators (<= v6).

For now, I would always use `rxjs/operators`

src/rules/effects/prefer-concat-latest-from.ts
timdeschryver: `rxjs/operators` 
We'll figure out how we want to use the rxjs imports later

tests/rules/prefer-concat-latest-from.test.ts
timdeschryver: Yep I forgot about it, but that's probably for the best 👍

tests/rules/prefer-concat-latest-from.test.ts
```diff
@@ -1,150 +1,276 @@
...
+class Test {
+  effectNOK3 = createEffect(() => {
+    return condition
+      ? this.actions$.pipe(
+          ofType(ProductDetailPage.loaded),
+          concatMap((action) =>
+            of(action).pipe(
+              concatLatestFrom(() => this.store.select$(something),), map( (one, other) => somethingElse()),
+            ),
+          ),
```
timdeschryver, line 242: This is fine, but the ideal scenario would be to also get rid of `concatMap` and `of`.
I'm afraid that this might lead to a lot more code/complexity though.
For now, let's leave it as is.

```suggestion
          concatLatestFrom(() => this.store.select$(something),), map( (one, other) => somethingElse()),
```

timdeschryver:
Shoot... merging via GitHub always messes up the linter... I'm on it

src/rules/effects/prefer-concat-latest-from.ts
timdeschryver: Reason: because that's what peoples are used to now, and we're sure that we don't break anything 😅

tests/rules/prefer-concat-latest-from.test.ts
rafaelss95: Yes, I thought about it while implementing and decided it wasn't worth the effort and we could also somehow break something, like for example if someone had structure inside the `concatMap`:

```ts
concatMap((action) => {
  performSideEffects();
  return of(action).pipe(withLatestFrom(...)))
})
```

src/rules/effects/prefer-concat-latest-from.ts
timdeschryver: 1. 👍
2. also a 👍, naming things is hard 😅

timdeschryver (approved):

PR #235, opened 2021-09-29
fix(no-dispatch-in-effects): remove fixer

### LINKED ISSUE #230: Bug fixer of no-dispatch-in-effects
## Behavior

The fixer introduced in #220 doesn't cover all cases and leads to incorrect behavior.

## Code snippet

In the following code snippet, the fixer removes the `dispatch` which is fine.

```ts
 map((books: Book[]) =>
              this.store.dispatch(BooksApiActions.searchSuccess({ books }))
),
```

But if we replace the `map` operator with the `tap` operator the fix is the same:

```ts
 tap((books: Book[]) =>
              (BooksApiActions.searchSuccess({ books }))
),
```

This changes the behavior because now, `BooksApiActions.searchSuccess` isn't dispatched.
I think this can become dangerous if used with the `--fix` flag.
Because of that reason, we should either make the fix smarter (is it possible to cover all cases?), or just to revert the change and implement it as a suggestion.

[ ] No

rafaelss95:
Fixes #230.


src/utils/selectors/index.ts
```diff
@@ -51,7 +51,7 @@ export const actionDispatch = (storeName: string) =>
 export const dispatchInEffects = (storeName: string) =>
   `${createEffectExpression} ${actionDispatch(
     storeName,
-  )} > MemberExpression` as const
+  )} > MemberExpression:has(Identifier[name='${storeName}'])` as const
```
rafaelss95, line 54: It was deleted by accident before 🙈

timdeschryver (approved):
Thanks for the quick fix!

PR #237, opened 2021-09-29
test: increase branch coverage

rafaelss95:
This PR increases branch coverage from 82.44% to 85.08%. It was mostly because we didn't have tests without `Store`.


src/utils/selectors/index.ts
```diff
@@ -68,6 +60,14 @@ export const pipeableSelect = (storeName: string) =>
 export const storeSelect = (storeName: string) =>
   `${storeExpression(storeName)}[callee.property.name='select']` as const
 
+export const storeDispatch = (storeName: string) =>
+  `${storeExpression(storeName)}[callee.property.name='dispatch']` as const
```
rafaelss95, line 64: I just noticed that we could reuse the `storeExpression` + I took the opportunity to rename this one to match the existing convention `store*`.

rafaelss95:
This is blocked by #236.

timdeschryver (approved):

PR #240, opened 2021-09-30
fix(no-multiple-actions-in-effects): use type-checking to report for all known nodes

rafaelss95:
Currently the rule only uses type-checking for explicit returns, but one could return a _computed_ `Array` using a implicit return (`ArrowFunctionExpression`) or `CallExpression`.


tests/rules/no-multiple-actions-in-effects.test.ts
```diff
@@ -96,112 +80,90 @@ ruleTester().run(path.parse(__filename).name, rule, {
...
     fromFixture(
       stripIndent`
+        @Injectable()
         export class Effects {
-          four$ = createEffect(() =>
+          effectNOK6$ = createEffect(() =>
             this.actions$.pipe(concatMap(() => {
               let actions: Action[] = [];
               return actions;
                      ~~~~~~~  [${messageId}]
```
rafaelss95, line 163: An interesting behavior I've noticed is that, this:

```
return actions.length > 0 ? actions : of(foo());
```

doesn't report as we may expect, **but** this:


```
return actions.length > 0 ? actions : null;
```

**does**.

---

That being said, to provide a full support for all possible cases like `?.`, `??`, `? :`, `||`, `&&` would increase the complexity and I'm not too sure it brings much value as no one complains until now.

tests/rules/no-multiple-actions-in-effects.test.ts
```diff
@@ -8,200 +8,171 @@ import { ruleTester } from '../utils'
...
-          eight$ = createEffect(() =>
+          readonly effectNOK4$ = createEffect(() =>
+            this.actions$.pipe(concatMapTo(actions))
+                                           ~~~~~~~  [${messageId}]
+          )
+        }`,
+    ),
+    fromFixture(
+      stripIndent`
+        const actions = [foo(), bar()]
```
rafaelss95, line 135: Just to ensure we report for functions that return `Arrays` as well:

```suggestion
        const actions = () => [foo(), bar()]
```

tests/rules/no-multiple-actions-in-effects.test.ts
```diff
@@ -8,200 +8,171 @@ import { ruleTester } from '../utils'
...
+        const actions = () => [foo(), bar()]
+        @Injectable()
+          export class Effects {
+          effectNOK5$ = createEffect(() =>
             this.actions$.pipe(
               exhaustMap(() => {
                 return of({}).pipe(
-                  switchMap(() => [foo(), bar()]),
-                                  ~~~~~~~~~~~~~~  [${messageId}]
+                  switchMap(() => actions),
```
rafaelss95, line 142:
```suggestion
                  switchMap(() => actions()),
```

tests/rules/no-multiple-actions-in-effects.test.ts
timdeschryver: Weird, I've taken a quick look but it's a mystery to me. 
 For now, let's keep it like thi

tests/rules/no-multiple-actions-in-effects.test.ts
```diff
@@ -8,200 +8,171 @@ import { ruleTester } from '../utils'
...
+      export class Effects {
+        effectOK2$ = createEffect(() =>
+          this.actions$.pipe(mapTo(action()))
+        )
+      }`,
+    `
+      @Injectable()
+      export class Effects {
+        effectOK3$ = createEffect(() =>
+          this.actions$.pipe(switchMapTop([foo()]))
```
timdeschryver, line 39: Note that these tests will fail every applying this change

```suggestion
          this.actions$.pipe(switchMapTo([foo()]))
```

tests/rules/no-multiple-actions-in-effects.test.ts
```diff
@@ -8,200 +8,171 @@ import { ruleTester } from '../utils'
...
+        )
+      }`,
+    `
+      @Injectable()
+      export class Effects {
+        effectOK4$ = createEffect(() =>
+          this.actions$.pipe(aconcatMapTo([bar()]))
```
timdeschryver, line 46:
```suggestion
          this.actions$.pipe(concatMapTo([bar()]))
```

tests/rules/no-multiple-actions-in-effects.test.ts
```diff
@@ -8,200 +8,171 @@ import { ruleTester } from '../utils'
...
+        @Injectable()
+          export class Effects {
+          effectNOK5$ = createEffect(() =>
             this.actions$.pipe(
               exhaustMap(() => {
                 return of({}).pipe(
-                  switchMap(() => [foo(), bar()]),
-                                  ~~~~~~~~~~~~~~  [${messageId}]
+                  switchMap(() => actions()),
+                                  ~~~~~~~  [${messageId}]
```
rafaelss95, line 143:
```suggestion
                                  ~~~~~~~~~  [${messageId}]
```

tests/rules/no-multiple-actions-in-effects.test.ts
rafaelss95: Yeah, I didn't dig too much into this, but maybe it's a problem with `couldBeType`.

timdeschryver: @rafaelss95 fyi, I updated the comment...  you're too fast for me 😂

tests/rules/no-multiple-actions-in-effects.test.ts
rafaelss95: I didn't change this part. I just converted the valid cases to be simple `strings` instead `objects`, because it was unnecessary.

I believe that these cases are like this to ensure we don't flag any custom rxjs operator, so it validates if your `RegExp` is restricted to built-in rxjs operators.

rafaelss95: Done! Check this out please.

rafaelss95: Worth mentioning that `rxjs` team plans to deprecate this xMapTo variants as stated here: https://github.com/ReactiveX/rxjs/issues/6399

rafaelss95: Worth mentioning that `rxjs` team plans to deprecate this `MapTo` variants as stated here: https://github.com/ReactiveX/rxjs/issues/6399.

tests/rules/no-multiple-actions-in-effects.test.ts
timdeschryver: 👀 it actually handles it correct - https://github.com/cartant/tsutils-etc/blob/master/source/could-be-type.ts#L30
The problem here, is that we should import `of` from `rxjs`.
Otherwise its type will just be `any` and thus isn't a union.

rafaelss95: I keep missing those imports within those strings 🤦

timdeschryver: You and me both...
Just a random thought, would it make sense to:

- create a seperate file for each case and divide them into a valid and invalid subfolder
- build the spec dynamically by reading those files and adding them as cases?

That way the individual file could be typechecked.

timdeschryver (approved):

tests/rules/no-multiple-actions-in-effects.test.ts
rafaelss95: You mean something like what [`tslint` used to do](https://github.com/palantir/tslint/blob/master/test/rules/adjacent-overload-signatures/ignore-accessors/test.ts.lint)? I think it's a fantastic idea. Just to clarify: would we use this approach only for rules with type-checking or...?

timdeschryver: @rafaelss95 I don't think we'll be able to pull it of straight away.
I didn't think of the messageId and the rules that have an output 😅

rafaelss95: Yes, that's not trivial on a second thought 😆

PR #241, opened 2021-09-30
fix(avoid-cyclic-effects): handle `PropertyDeclaration` properly

### LINKED ISSUE #223: Bug ngrx/avoid-cyclic-effects
## Behavior
I don't see any cyclic issue here but I am getting this rule violation from eslint.
I am not sure how internally it is evaluated such an effect like this. Maybe just a naming issue?

## Code snippet

```
    public postPortInData$ = createEffect(() =>
        this.actions$.pipe(
            ofType<PostPortInData>(OrderEntityActionTypes.postPortInData),
            map((action) => action.payload),
            concatLatestFrom(() => this.appStoreFacadeService.order.orderEntity$),
            switchMap(([orderContext, orderEntity]) =>
                this.orderEntityService.queueContextCall(orderContext, orderEntity).pipe(
                    map(() => new PostPortInDataSuccess()),
                    catchError(() => of(new ConfigErrorShowModal()))
                )
            )
        )
    );
```

[x] No

rafaelss95:
Fixes #223.

### DIFF
--- a/tests/rules/avoid-cyclic-effects.test.ts
+++ b/tests/rules/avoid-cyclic-effects.test.ts
@@ -122,79 +122,34 @@ ruleTester().run(path.parse(__filename).name, rule, {
     // https://github.com/timdeschryver/eslint-plugin-ngrx/issues/223
     `
       import { Injectable } from '@angular/core';
-      import { Actions, createEffect, concatLatestFrom, ofType } from '@ngrx/effects';
+      import { Actions, createEffect, ofType } from '@ngrx/effects';
       import { Action } from '@ngrx/store';
       import { of } from 'rxjs';
-      import { catchError, map, switchMap } from 'rxjs/operators';
+      import { switchMap } from 'rxjs/operators';
 
       enum OrderEntityActionTypes {
         postPortInData = '[Order Entity] Post PortIn Data',
         postPortInDataSuccess = '[Order Entity] Post PortIn Data Success',
-        postPartnerData = '[Order Entity] Post provider Data',
       }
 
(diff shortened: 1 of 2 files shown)


tests/rules/avoid-cyclic-effects.test.ts
```diff
@@ -118,7 +118,85 @@ ruleTester().run(path.parse(__filename).name, rule, {
           private actions$: Actions,
         ) {}
       }
-    `),
+    `,
+    // https://github.com/timdeschryver/eslint-plugin-ngrx/issues/223
+    `
+      import { Injectable } from '@angular/core';
```
timdeschryver, line 124: Could we make the setup here simpler, or do you think it's worth to keep the full example?
I think it's easier to read if we can get rid of the service and some other boilerplate. something like this:

```ts
enum OrderEntityActionTypes {
        postPortInData = '[Order Entity] Post PortIn Data',
        postPortInDataSuccess = '[Order Entity] Post PortIn Data Success',
      }
      class PostPortInData implements Action {
        public readonly type = OrderEntityActionTypes.postPortInData
      }
      class PostPortInDataSuccess implements Action {
        public readonly type = OrderEntityActionTypes.postPortInDataSuccess
      }

        public postPortInData$ = createEffect(() =>
          this.actions$.pipe(
            ofType<PostPortInData>(OrderEntityActionTypes.postPortInData),
            switchMap(() => of(new PostPortInDataSuccess())),
          ),
        )
```

timdeschryver (approved):

PR #243, opened 2021-10-01
fix(prefer-effect-callback-in-block-statement): handle all kinds of `createEffect` properly

rafaelss95:
Part of #242, this fixes the following issues:

- [x] handle `createEffect` in any place, not only in `ClassProperty`;
- [x] handle parametrized `createEffect`;
- [x] correct the fix if `createEffect#body` is parenthesized. Currently `createEffect(() => (...)` become `createEffect(() => ({ ... })`.

### DIFF
--- a/src/rules/effects/prefer-effect-callback-in-block-statement.ts
+++ b/src/rules/effects/prefer-effect-callback-in-block-statement.ts
@@ -1,7 +1,7 @@
-import type { TSESTree } from '@typescript-eslint/experimental-utils'
-import { ESLintUtils } from '@typescript-eslint/experimental-utils'
+import type { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils'
+import { ASTUtils, ESLintUtils } from '@typescript-eslint/experimental-utils'
 import path from 'path'
-import { createEffectFunction, docsUrl } from '../../utils'
+import { createEffectExpression, docsUrl } from '../../utils'
 
 export const messageId = 'preferEffectCallbackInBlockStatement'
 export type MessageIds = typeof messageId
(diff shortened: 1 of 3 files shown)


src/rules/effects/prefer-effect-callback-in-block-statement.ts
```diff
@@ -26,19 +26,52 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   },
   defaultOptions: [],
   create: (context) => {
+    const sourceCode = context.getSourceCode()
+    const nonParametrizedEffect =
+      `${createEffectExpression} > ArrowFunctionExpression > .body[type!=/^(ArrowFunctionExpression|BlockStatement)$/]` as const
+    const parametrizedEffect =
+      `${createEffectExpression} > ArrowFunctionExpression > ArrowFunctionExpression > .body[type!='BlockStatement']` as const
+    const parametrizedEffectWithinBlockStatement =
```
rafaelss95, line 34: I've tried to be as explicit as possible in the variables, but feel free to suggest if that's the case.

src/rules/effects/prefer-effect-callback-in-block-statement.ts
```diff
@@ -26,19 +26,52 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+      `${createEffectExpression} > ArrowFunctionExpression > BlockStatement > ReturnStatement > ArrowFunctionExpression > .body[type!='BlockStatement']` as const
+
     return {
-      [`${createEffectFunction} > CallExpression`](
-        node: TSESTree.CallExpression,
+      [`${nonParametrizedEffect}, ${parametrizedEffect}, ${parametrizedEffectWithinBlockStatement}`](
```
rafaelss95, line 38: I tried `${createEffectExpression} > :matches(${nonParametrizedEffect}, ${parametrizedEffect}, ${parametrizedEffectWithinBlockStatement})` to avoid repeating the `createEffectExpression`, but for some reason it didn't work 🤷‍♂️

timdeschryver (approved):

PR #245, opened 2021-10-01
fix(prefer-concat-latest-from): disable pre ngrx v12

timdeschryver:
Thought: we could also use these to disable package rules if the package isn't installed.


package.json
```diff
@@ -53,7 +53,8 @@
     "@angular-devkit/schematics": "^12.2.2",
     "@typescript-eslint/experimental-utils": "^4.31.1",
     "eslint-etc": "^4.2.4",
-    "eslint-plugin-rxjs": "^3.3.6"
+    "eslint-plugin-rxjs": "^3.3.6",
+    "semver": "^7.3.5"
   },
   "peerDependencies": {
```
rafaelss95, line 59: Let me know your opinion on upgrade those `peerDependencies` + adding `@ngrx` + `rxjs`..

src/rules/effects/prefer-concat-latest-from.ts
```diff
@@ -33,6 +36,10 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
   },
   defaultOptions: [],
   create: (context) => {
+    if (!ruleEnabled) {
```
rafaelss95, line 39: 1. Would it be possible to us to test this behavior? Maybe mocking the `require('@ngrx/effects/schematics-core')`?
2. What about `console.warn` (or something like that) informing the user that the rule isn't supported given XX version of `@ngrx`?

package.json
rafaelss95, line 59: Let me know your opinion on upgrade those `peerDependencies` + adding `@ngrx` + `rxjs`..

src/rules/effects/prefer-concat-latest-from.ts
rafaelss95: 1. Would it be possible to us to test this behavior? Maybe mocking the `require('@ngrx/effects/schematics-core')`?
2. What about `console.warn` (or something like that) informing the user that the rule isn't supported given XX version of `@ngrx`?

src/utils/helper-functions/versions.ts
```diff
@@ -0,0 +1,23 @@
+import * as semver from 'semver'
```
rafaelss95, line 1:
```suggestion
import semver from 'semver'
```

src/utils/helper-functions/versions.ts
```diff
@@ -0,0 +1,23 @@
...
+let ngrxEffectsVersion: string | undefined
+function getNgrxEffectsVersion(): string {
+  if (ngrxEffectsVersion) return ngrxEffectsVersion
+
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const effects = require('@ngrx/effects/schematics-core') as {
+      platformVersion: string
+    }
+    // version in file is ^x.x.x, but this isn't valid for semver
```
rafaelss95, line 12: Really? I was pretty sure that it worked without any problems...

src/utils/helper-functions/versions.ts
```diff
@@ -0,0 +1,23 @@
+import * as semver from 'semver'
+
+let ngrxEffectsVersion: string | undefined
+function getNgrxEffectsVersion(): string {
+  if (ngrxEffectsVersion) return ngrxEffectsVersion
+
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const effects = require('@ngrx/effects/schematics-core') as {
+      platformVersion: string
```
rafaelss95, line 10: Is it optional? If so:

```suggestion
      platformVersion?: string
```

If not, let's remove the `optional chaining` below.

src/utils/helper-functions/versions.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+    // version in file is ^x.x.x, but this isn't valid for semver
+    ngrxEffectsVersion = effects.platformVersion?.substr(1)
+  } catch (err) {
```
rafaelss95, line 14:
```suggestion
  } catch {
```

src/utils/helper-functions/versions.ts
```diff
@@ -0,0 +1,23 @@
...
+    ngrxEffectsVersion = '0.0.0'
+  }
+
+  return ngrxEffectsVersion ?? '0.0.0'
+}
+
+export function ngrxEffectsVersionSatisfies(version: string) {
+  return semver.satisfies(getNgrxEffectsVersion(), version)
+}
```
rafaelss95, line 23: Instead of _memoize_ only the `ngrxEffectsVersion`, I'd _memoize_ this whole function (and remove the `ngrxEffectsVersion` of the top):

```suggestion
let effectsVersionSatisfies: boolean | undefined
export function ngrxEffectsVersionSatisfies(version: string): boolean {
  return (
    effectsVersionSatisfies ??
    (effectsVersionSatisfies = semver.satisfies(
      getNgrxEffectsVersion(),
      version,
    ))
  )
}
```

package.json
timdeschryver: Oh yea, I forgot...
I don't think that's a good idea, this plugin covers 3 separate ngrx packages.
It's perfectly fine to not have one of them installed.
Maybe with optional dependencies like [this](https://github.com/timdeschryver/ngrx-immer/blob/main/src/package.json#L31) ? As long as it isn't needed, I wouldn't add them though.

src/utils/helper-functions/versions.ts
timdeschryver: Yea, it is - the ngrx version is scripted and the caret will be there. 
At least for now 😅, I'll add a check for it to be sure.

src/utils/helper-functions/versions.ts
timdeschryver: It's not optional in the last few versions of ngrx, before that, it wasn't exported 😅
That's why I wanted to be cautious

src/utils/helper-functions/versions.ts
timdeschryver: I think that this won't work if we have different rules working on different verion requirements?

rafaelss95: You got a point 😅.

What about this update to `ngrxEffectsVersionSatisfies`?

```ts
const effectsVersionSatisfies: Record<string, boolean> = {}
export function ngrxEffectsVersionSatisfies(version: string) {
  return (
    effectsVersionSatisfies[version] ??
    (effectsVersionSatisfies[version] = semver.satisfies(getVersion(), version))
  )
}

```

src/utils/helper-functions/versions.ts
rafaelss95: Aha, in that case, I'd suggest the first option (marking it with `?`).

timdeschryver:
> Would it be possible to us to test this behavior? Maybe mocking the require('@ngrx/effects/schematics-core')?

That was the idea, but I was in a hurry.
I think it can be achieved with creating a new test file, let me try

> What about console.warn (or something like that) informing the user that the rule isn't supported given XX version of @ngrx?

I'm not a big fan of it. 
It will just clutter the output, and it isn't applicable to the user since she/he is on a different version.
I was thinking to add it to the docs though.

package.json
rafaelss95: You have a point about `@ngrx/*` stuff... let's not add them so.

src/utils/helper-functions/versions.ts
timdeschryver: Eh, I added it but I had some struggles to make it work during tests.
I thought each test file would have been isolated but it seems like that's not the case.
Thus, it required a bit more setup - let me know what you think

timdeschryver:
And what do you think of a6c9c8ccb22e539281fedf30e35250ac5e608561 - in this commit I made the check on version generic so it can be used for every rule that uses the custom rule creator. 
Maybe an overkill for now or forever?

timdeschryver:
Extra bonus points to make it generic so it can generate our docs? 🤔

src/rule-creator.ts
```diff
@@ -0,0 +1,51 @@
+import type { TSESLint } from '@typescript-eslint/experimental-utils'
+import { ESLintUtils } from '@typescript-eslint/experimental-utils'
+import type { NGRX_MODULE } from './utils'
+import { docsUrl, MODULE_PATHS, ngrxVersionSatisfies } from './utils'
+
+export type NgRxRuleModule<TMessageIds extends string = string> =
+  TSESLint.RuleModule<TMessageIds, unknown[]> & {
+    meta: { module: NGRX_MODULE; version?: string }
```
rafaelss95, line 8: 1. I'd prefer not to call this `module`, as `module` already has meaning in the JS world and can be problematic in case of destructuring, for example.

Some suggestions are: `moduleName`, `ngrxModule` or `ngrxModuleName`.

2. Why is it required? I'm asking because I was imagining that it would be great if we changed all the rules to use this _new way_ (of course, the change would be in another PR), thus keeping all the rules using a single implementation.

src/utils/helper-functions/versions.ts
```diff
@@ -0,0 +1,55 @@
+import semver from 'semver'
+
+const noopVersion = '0.0.0'
+const versionsCache: Map<string, string> = new Map()
+const satisfiesCache: Map<string, boolean> = new Map()
```
rafaelss95, line 5: nit:

```suggestion
const versionsCache = new Map<string, string>()
const satisfiesCache = new Map<string, boolean>()
```

tests/rules/prefer-concat-latest-from.test.ts
```diff
@@ -118,3 +119,26 @@ ruleTester().run(path.parse(__filename).name, rule, {
...
+    valid: [
+      `
+      export class Effect {
+        effect$ = createEffect(
+          () =>
+            this.actions$.pipe(
+              ofType(CollectionApiActions.addBookSuccess),
+              withLatestFrom(action => this.store.select(fromBooks.getCollectionBookIds)),
+              switchMap(([action, bookCollection]) => {
+                return {}
```
rafaelss95, line 136:
```suggestion
                return of({})
```

src/utils/helper-functions/versions.ts
rafaelss95: Aha, I missed that by default `caughterrors` is set to `none`: https://eslint.org/docs/rules/no-unused-vars#caughterrors

rafaelss95:
> And what do you think of [a6c9c8c](https://github.com/timdeschryver/eslint-plugin-ngrx/commit/a6c9c8ccb22e539281fedf30e35250ac5e608561) - in this commit I made the check on version generic so it can be used for every rule that uses the custom rule creator. Maybe an overkill for now or forever?

Can you imagine I had a very similar idea earlier today? Hahaha, what a coincidence! Mine idea was patching the `ESLint.RuleCreator` adding a `isEnabled()` predicate, similar to the old `TSLint`, but ofc your idea is not only cleaner, but more robust. 🏆 

I don't think this is an overkill: the implementation isn't that difficult, and I imagine that we'll be using it more and more as `@ngrx` evolves over time, deprecating/adding features.

> Extra bonus points to make it generic so it can generate our docs? 🤔 [4ee45fe](https://github.com/timdeschryver/eslint-plugin-ngrx/commit/4ee45fecd18ee50221f3950cca8a813430643b54)

That's another great point ofc!

src/utils/helper-functions/versions.ts
```diff
@@ -0,0 +1,55 @@
...
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const ngrxVersion = require(`${pkg}/schematics-core`) as {
+      platformVersion?: string
+    }
+    const version = ngrxVersion.platformVersion
+      ?.split('')
+      .filter((c) => (c >= '0' && c <= '9') || c === '.')
+      .join('')
+    return version
```
rafaelss95, line 26: If I'm understanding it correctly, this strips all non-numeric + dots... what about:

```suggestion
    return ngrxVersion.platformVersion?.replace(/[^\d\.]/g, '')
```

src/rule-creator.ts
timdeschryver: 1. You're right, I'm changing it to `ngrxModule `
2. It's currently set in one of the config scripts, and it's set based on the folder of the rule. I have kept it this way in order to change too much. In the future we can use `createRule` for all of our rules, in those case I would expect that `ngrxModule` is required because we rely on it to build specific configs. After all rules are "refactored", we can remove the code that sets the module atomically based on the folder.

timdeschryver:
The comments should be resolved

rafaelss95 (approved):
LGTM 🥳

PR #247, opened 2021-10-02
refactor: improve selectors

rafaelss95:
This PR refactor some selectors to use `.{property}` so we don't need to destruct or create union types to represent it properly.


docs/rules/use-selector-in-select.md
```diff
@@ -1,6 +1,6 @@
 # Use Selector In Select
 
-> Using a selector in a select method is preferred in favor of strings or props drilling.
+> Using a selector in a select method is preferred over string or props drilling.
```
rafaelss95, line 3: I was copying (😅) this description to #248 and realized that "over" fits better than "in favor of" in this case, so I decided to change this too. Let me know if it's fine for you too.

src/rules/store/use-selector-in-select.ts
```diff
@@ -23,13 +23,13 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
     docs: {
       category: 'Best Practices',
       description:
-        'Using a selector in a select method is preferred in favor of strings or props drilling.',
+        'Using a selector in a `select` method is preferred over `string` or `props drilling`.',
```
rafaelss95, line 26: I was copying (😅) this description to #248 and realized that "over" fits better than "in favor of" in this case, so I decided to change this here as well. Let me know if it's fine for you too.

timdeschryver: Yep, this is better 👍

timdeschryver (approved):

Issue #238, opened 2021-09-29
A rule to enforce `.dispatch(action())`


rafaelss95:
## Information about the new rule
As mentioned in https://github.com/timdeschryver/eslint-plugin-ngrx/pull/211#discussion_r712270311, we could create a rule for this.

Do you have any name in mind?

## A snippet of incorrect code
```ts
this.store$.dispatch({ type: 'myAction' });
this.store$.dispatch(new MyAction());

```

## A snippet of correct code
```ts
this.store$.dispatch(action());
```
## I would be willing to submit a PR for the docs :heart:

[x] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

timdeschryver:
What about `use-actions-in-dispatch` ?

rafaelss95:
My only concern to this name is that `this.store$.dispatch(new MyAction())` would be flagged as well.. so imagine users asking theirselves: "wait, isn't `new MyAction()` an action? Is this a incorrect report?" 🤔

Maybe something like `use-action-creator-in-dispatch`?

timdeschryver:
Works for me 👍

PR #248, opened 2021-10-02
feat(prefer-action-creator-in-of-type): add rule

### LINKED ISSUE #239: A rule to enforce `ofType(NON_STRING)`
As mentioned in https://github.com/timdeschryver/eslint-plugin-ngrx/pull/211#discussion_r711401470, we could create a rule for this.

Do you have any name in mind?
- `no-of-type-literal`?
- `use-actions-in-of-type`? (taking the idea of `use-selector-in-select`)
- `prefer-actions-in-of-type`?
- other...

```ts
this.actions$.pipe(ofType('myAction'));
```

## A snippet of correct code
```ts
this.actions$.pipe(ofType(myAction));
```

[ ] No

rafaelss95:
Closes #239.


src/rules/effects/use-action-creator-in-of-type.ts
```diff
@@ -0,0 +1,36 @@
...
+import path from 'path'
+import { docsUrl } from '../../utils'
+
+export const messageId = 'noOfTypeLiteral'
+export type MessageIds = typeof messageId
+
+type Options = []
+
+export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
+  name: path.parse(__filename).name,
```
rafaelss95, line 12: I went with `use-action-creator-in-of-type` to conform to the `use-action-creator-in-dispatch` as per discussion in https://github.com/timdeschryver/eslint-plugin-ngrx/issues/238#issuecomment-931360802.

src/rules/effects/use-action-creator-in-of-type.ts
```diff
@@ -0,0 +1,36 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      category: 'Best Practices',
+      description: 'Using action in `ofType` is preferred over string.',
+      recommended: 'warn',
```
rafaelss95, line 18: Not sure about this recommendation...

src/rules/effects/use-action-creator-in-of-type.ts
rafaelss95: After thinking twice, I realized that `prefer-*` seems to be fits better in this context and the other in #249... so I'd prefer `prefer-action-creator-in-of-type` and `prefer-action-creator-in-dispatch`.

---

From here on are just random thoughts/suggestions that involve breaking changes and that don't involve this PR...

About the existing rules with the prefix `use-*` (`use-selector-in-select` and `use-consistent-global-store`), I'd suggest to rename these to `prefer-selector-in-select` (for the same reason mentioned above) and `consistent-global-store`(it seems that starting with `consistent-*` is a standard  for this kind of rule... check the `consistent-*` rules [here](https://github.com/eslint/eslint/tree/master/lib/rules) and [here](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin/src/rules)).

On other hand, I guess the `use-effects-lifecycle-interface` makes sense to keep as is.

rafaelss95:
Converting to draft until we reach consensus on name...

src/rules/effects/use-action-creator-in-of-type.ts
timdeschryver: +1 for `prefer`.
The suggestion to rename the existing rule is indeed better.
I'll create a new issue; we can do it later when we've stabilized.
Maybe that other breaking changes will come to light.

src/rules/effects/use-action-creator-in-of-type.ts
timdeschryver: This is good

tests/rules/prefer-action-creator-in-of-type.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    @Injectable()
+    class Test {
+      effectOK = createEffect(() => this.actions$.pipe(ofType(userActions.ping)))
+
+      constructor(private readonly actions$: Actions) {}
+    }`,
+    `
+    @Injectable()
+    class Test {
+      effectOK1 = createEffect(() => this.actions$.pipe(ofType(userActions.ping.type)))
```
rafaelss95, line 21: Redundant, but indeed using `action creators` 🤷‍♂️

docs/rules/prefer-action-creator-in-of-type.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+effectOK = createEffect(() => this.actions$.pipe(ofType(userActions.ping)))
+
+effectOK1 = createEffect(() =>
+  this.actions$.pipe(ofType(userActions.ping.type)),
+)
```
timdeschryver, line 24: While this is technically correct, I would prefer to leave it out because I'm afraid that it might be confusing?

```suggestion
```

timdeschryver (approved):

PR #249, opened 2021-10-02
feat(prefer-action-creator-in-dispatch): add rule

### LINKED ISSUE #238: A rule to enforce `.dispatch(action())`

rafaelss95:
Closes #238.


rafaelss95:
Converting to draft until we reach consensus on name...

src/rules/store/prefer-action-creator-in-dispatch.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const messageId = 'preferActionCreatorInDispatch'
+export type MessageIds = typeof messageId
+
+type Options = []
+
+export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
+  name: path.parse(__filename).name,
+  meta: {
+    type: 'problem',
```
timdeschryver, line 14: Should we use `suggestion` for all of our `prefer-` rules 🤔

rafaelss95: Addressed 💯. Your comment made me realize that today we have `prefer-concat-latest-from` as `type: 'suggestion'`, but shouldn't it be `problem` instead, as misusing `withLatestFrom` can lead to some problems?

timdeschryver: You're also right! That should be `problem`.
This is hard 😅

timdeschryver (approved):

PR #250, opened 2021-10-02
feat(prefer-action-creator): add rule

rafaelss95:
As discussed in https://github.com/timdeschryver/eslint-plugin-ngrx/discussions/134#discussioncomment-1410777.


tests/rules/good-action-hygiene.test.ts
```diff
@@ -7,10 +7,10 @@ import { ruleTester } from '../utils'
 ruleTester().run(path.parse(__filename).name, rule, {
   valid: [
     `export const loadCustomer = createAction('[Customer Page] Load Customer')`,
-    `export const loadCustomer = createAction('[Customer API] Load Customer Success, props<{ customer: Customer }>()')`,
-    `export const loadCustomer = createAction('[Customer API] Load Customer Failed, (error: string) => ({ error, timestamp: +Date.now()})')`,
-    `export const loadCustomer = createAction(iDoNotCrash)`,
-    `export const loadCustomer = createActionType('Just testing')`,
+    `export const loadCustomerSuccess = createAction('[Customer API] Load Customer Success', props<{ customer: Customer }>())`,
+    `export const loadCustomerFail = createAction('[Customer API] Load Customer Fail', (error: string) => ({ error, timestamp: +Date.now() }))`,
```
rafaelss95, line 11: The quotes were closing in the wrong place here.

src/rules/store/prefer-action-creator.ts
```diff
@@ -0,0 +1,40 @@
...
+    schema: [],
+    messages: {
+      [messageId]:
+        'Using `Action class` is not preferred, use an `action creator` instead.',
+    },
+  },
+  defaultOptions: [],
+  create: (context) => {
+    return {
+      [`ClassDeclaration:has(TSClassImplements:matches([expression.name='Action'], [expression.property.name='Action']))`](
```
rafaelss95, line 30: Do you see any problem looking for `Action` like this? I'm asking because I have the impression that the `Action` is a very trivial naming and may not always be the one from `@ngrx/store`... on the other hand, we have many other rules where we look for various methods/functions no matter where it comes from, that I don't think having more of this would be a problem.

timdeschryver: I think this is fine.
For extra safety we could also verify that the class has a `type` property?

timdeschryver (approved):
When the conflict is resolved we can merge this in

PR #254, opened 2021-10-04
fix(avoid-cyclic-effects): correctly handle non `ParameterProperties` and computed `dispatch`

rafaelss95:
- [x] change type from `suggestion` to `problem`;
- [x] correctly handle non `ParameterProperty` `actions`;
- [x] uses type-checking to correctly handle computed `effectConfig`;
- [x] adds more tests (without `actions`, with non `ParameterProperty` `actions`, with various types passed to `ofType`, returning multiple actions, various combinations of `dispatch`, conditional with multiple reports, etc.).

**Before**:
Before

**After**:
After


tests/rules/avoid-cyclic-effects.test.ts
```diff
@@ -234,19 +296,57 @@ ruleTester().run(path.parse(__filename).name, rule, {
             this.actions$.pipe(
             ~~~~~~~~~~~~~~~~~~ [${messageId}]
               ofType(genericFoo),
+              mergeMapTo(of(genericFoo()))
```
rafaelss95, line 299: Could give me a hand here please? I have no idea why this test if failing (and yes, I checked the imports 😅).

Note that `mergeMapTo([...])` and `mergeMap(() => somethingElse)` are reported as expected, but this one here isn't 🤔

src/rules/effects/avoid-cyclic-effects.ts
```diff
@@ -115,38 +116,67 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
+    const pipeNodes = new Set<TSESTree.CallExpression>()
+
     return {
-      [`${createEffectExpression}:not([arguments.1]:has(Property[key.name='dispatch'][value.value=false])) CallExpression[callee.property.name='pipe'][callee.object.property.name='${actionsName}']`]:
-        checkNode,
+      [createEffectExpression]({
+        arguments: [, effectConfig],
+      }: TSESTree.CallExpression) {
+        if (!effectConfig) {
+          return (hasDispatchFalse = false)
```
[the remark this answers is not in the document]
timdeschryver: I prefer the two-liner, "one-liner tricks" can be tricky sometimes 

```ts
hasDispatchFalse = false
return
```

tests/rules/avoid-cyclic-effects.test.ts
timdeschryver: It has something to do with the effect config parameter, if I set it to `{dispatch:true}` the test succeeds.
If I change another test, and remove that parameter that test also fails, e.g.

```diff
@Injectable()
      class Effect {
        foo$: CreateEffectMetadata
        constructor(customActions$: Actions) {
          this.foo$ = createEffect(() =>
            customActions$.pipe(
            ~~~~~~~~~~~~~~~~~~~ [${messageId}]
              ofType(genericFoo),
              switchMapTo([genericBar(), genericFoo()]),
            ),
-            { useEffectsErrorHandler: true }
+         configDispatchTrue // or don't provide a config and see it fail too
          )
        }
      }
```

timdeschryver: Heh... this is strange.
I have to go now but I'll take another look later/tomorrow.

src/rules/effects/avoid-cyclic-effects.ts
```diff
@@ -115,38 +116,67 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
...
-      // e.g. const genericFoo = createAction(`${subject} FOO`); (resolves to 'string')
-      if (typeChecker.typeToString(actionType) === 'string') {
-        return []
-      }
-      return [typeChecker.typeToString(actionType)]
+      return [typeChecker.typeToString(symbolType)]
     }
 
+    let hasDispatchFalse = false
+    const pipeNodes = new Set<TSESTree.CallExpression>()
```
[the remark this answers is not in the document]
rafaelss95: Do you have a suggestion? 😄

src/rules/effects/avoid-cyclic-effects.ts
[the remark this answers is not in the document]
rafaelss95: Nice. Just added the rule [`no-return-assign`](https://eslint.org/docs/rules/no-return-assign) to keep this behavior consistent.

tests/rules/avoid-cyclic-effects.test.ts
timdeschryver: Yep, it's something iffy...
I didn't had the time to take a close look, I'll try to do it this week though.

PR #256, opened 2021-10-04
fix(prefer-action-creator-in-dispatch): properly handle actions with literal objects

### LINKED ISSUE #255: Bug prefer-action-creator-in-dispatch
## Behavior

I'm dispatching an action that takes a props object, which seems to follow the spirit of this rule -- it uses an action creator. Maybe the rule sees the props object in there and thinks it's an action object? It's also possible I could be writing this better :)

## Code snippet

Problem line that trips the rule:
`this.store.dispatch(Auth.Login({ payload: $event }));`

Underlying action imported from another file:
`export const Login = createAction('[Auth] Login', props<{ payload: Authenticate }>());`

[X] No

rafaelss95:
Fixes #255.

### DIFF
--- a/src/rules/store/prefer-action-creator-in-dispatch.ts
+++ b/src/rules/store/prefer-action-creator-in-dispatch.ts
@@ -1,7 +1,14 @@
 import type { TSESTree } from '@typescript-eslint/experimental-utils'
 import { ESLintUtils } from '@typescript-eslint/experimental-utils'
 import path from 'path'
-import { docsUrl, findNgRxStoreName, storeDispatch } from '../../utils'
+import {
+  docsUrl,
+  findNgRxStoreName,
+  getNearestUpperNodeFrom,
+  isCallExpression,
+  isCallExpressionWith,
+  storeDispatch,
+} from '../../utils'
 
 export const messageId = 'preferActionCreatorInDispatch'
 export type MessageIds = typeof messageId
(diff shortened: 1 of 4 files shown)


src/rules/store/prefer-action-creator-in-dispatch.ts
```diff
@@ -33,6 +40,22 @@ export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
       [`${storeDispatch(storeName)} :matches(NewExpression, ObjectExpression)`](
```
rafaelss95, line 40: IMHO, this fix is far from desired, but we can't simply do `${storeDispatch(storeName)} > :matches(NewExpression, ObjectExpression)` (note the `>`) because by doing so, we would miss several possible cases, like:

```ts
this.store$.dispatch(useObject ? { type: 'custom' } : new CustomAction())
this.store$.dispatch(somethingNullable ?? { type: 'custom' })
```

and much more... the expressions aren't necessarily the `first-child`.

---

I am increasingly convinced that the correct approach to getting `constructor` instances is the one I submitted at https://github.com/timdeschryver/eslint-plugin-ngrx/pull/221, which, in addition to looking more performatic, since we would query `store$`/`actions$/`whater` only once and look for its references instead of two (or in this case, **three**), it is also less prone to false positives, as we **do** know that the reference is in fact accurate.

src/utils/helper-functions/guards.ts
```diff
@@ -54,3 +55,21 @@ export function isIdentifierOrMemberExpression(
 export function isTypeReference(type: ts.Type): type is ts.TypeReference {
   return type.hasOwnProperty('target')
 }
+
+export function isCallExpressionWith(
```
rafaelss95, line 59: Not sure about the name + the place to put it.

src/rules/store/prefer-action-creator-in-dispatch.ts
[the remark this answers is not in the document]
timdeschryver: I don't really see how #211  could help here, but let's first merge that one - and then I'm sure it will click here once I see it

src/utils/helper-functions/guards.ts
timdeschryver: Either here or in utils.
Let's leave it here, if we need it somewhere else we can move it

timdeschryver (approved):
I'm going to merge this in, since this will fix #255.
We can refactor after #221 lands.

src/rules/store/prefer-action-creator-in-dispatch.ts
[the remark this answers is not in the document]
rafaelss95: Hmm, you may be right. I was thinking about something like comparing scopes, but only in theory... I don't know if it's so trivial in practice. I'll try to see what I can get later.

PR #158, opened 2021-07-15
refactor: remove `dist` import references

rafaelss95:
A small change that replaces `dist` references on import statement.

PR #257, opened 2021-10-05
test: use uvu to run rule tests for better reports

timdeschryver:
Currently, running the rule tests is not friendly because it abruptly stops when a rule test fails.
It also doesn't show why it fails.

In this PR, we're going to resolve this because it can waste some time and it was on my todo list 😅

The trick is to run the tests in an uvu context, with it uvu will:

- gracefully handle exceptions and continue to run other tests
- show the exception message (formatted) so we can clearly see what's wrong

As a fix, we got two options and I would like to hear your opinion @rafaelss95 before I migrate the other rules.

We could either do it like `avoid-combining-selectors`, and export the tests (we should then also rename them to fixtures to make it clear) and then iterate over them, and create a test for each fixture. This gives us a good overview (name of the fixture, and the formatted exception).

```txt
  FAIL  "avoid-combining-selectors.fixture"
    Error endColumn should be 89 (90 strictEqual 89)  (strictEqual)

        ++89    (Expected)
        --90    (Actual)

```

The other option is to keep the rules test in place but wrap them in an uvu test, like did for `avoid-cyclic-effects`

```


rules\avoid-cyclic-effects.test.ts
✘   (0 / 1)

   FAIL  "C:\Users\tdeschryver\dev\libs\eslint-plugin-ngrx\tests\rules\avoid-cyclic-effects.test.ts"
    Error endColumn should be 30 (29 strictEqual 30)  (strictEqual)

        ++30    (Expected)
        --29    (Actual)
```


timdeschryver:
That's a great idea!
I'm afraid that if I convert all rules, that this will conflict with #221 though.
If  it's OK I prefer to wait until that one is merged and then I'll update all tests.
If the "refactored" version is helpful for #221, feel free to update the rules that you refactor in #221.

tests/rules/avoid-duplicate-actions-in-reducer.test.ts
```diff
@@ -1,176 +1,183 @@
+import type { InvalidTestCase } from '@typescript-eslint/experimental-utils/dist/ts-eslint'
```
rafaelss95, line 1:
```suggestion
import type { TSESLint } from '@typescript-eslint/experimental-utils'
```

We definetly should use the [`@typescript-eslint/no-restricted-imports`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-restricted-imports.md) (available since [`@typescript-eslint@4.32.0`](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v4.32.0)) to make sure we don't commit this mistake again as it's a common one (references I've been involved in 😄: https://github.com/timdeschryver/eslint-plugin-ngrx/pull/158, https://github.com/angular-eslint/angular-eslint/pull/260#discussion_r547471168, https://github.com/typescript-eslint/typescript-eslint/issues/2931 and https://github.com/angular-eslint/angular-eslint/pull/711).

tests/rules/avoid-duplicate-actions-in-reducer.test.ts
```diff
@@ -1,176 +1,183 @@
...
+    on(def, state => state),
+    on(ghi, state => state),
+  )`,
+  // does not crash when no arguments present
+  `
+  export const reducer = createReducer(
+    {},
+    on(),
+  )`,
+]
```
rafaelss95, line 40: I'd add this where possible:

```suggestion
] as const
```

tests/rules/avoid-duplicate-actions-in-reducer.test.ts
```diff
@@ -1,176 +1,183 @@
...
+
+const invalid: InvalidTestCase<MessageIds, []>[] = [
```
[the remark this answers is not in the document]
rafaelss95: Interesting.. the `as const` should be sufficient here, however [`suggestions` doesn't accept a `readonly []`](https://github.com/typescript-eslint/typescript-eslint/blob/101abc72f0baced5cdf609d99418ad9f10830c73/packages/experimental-utils/src/ts-eslint/RuleTester.ts#L111) so we still can't use it right now.

For now, unfortunately, I think that's what we have as an option. Once my PR (https://github.com/typescript-eslint/typescript-eslint/pull/3844) has been reviewed/merged, we will be able to switch it to `as const` (this will only be possible in v2 though - when we update to `@typescript-eslint@v5`).

I'd just change it to `const invalid: Invalid TestCase <string, []>[] = [` because besides being a temporary solution, we don't really need to import `MessageIds` as it would give error in `ruleTester.run` if the type is different. Also, we're removing the `export` from `MessageIds` in #259 😄

tests/rules/avoid-duplicate-actions-in-reducer.test.ts
rafaelss95: I just realized that `@typescript-eslint/no-restricted-imports` won't help us right now as there's a bug there that it was not reporting `type-only` imports properly (I sent a small fix there though, PR: https://github.com/typescript-eslint/typescript-eslint/pull/3996).

For now, we could use [**`no-restricted-imports`**](https://eslint.org/docs/rules/no-restricted-imports) instead and change it when updating to `@typescript-eslint@v5`.

tests/rules/avoid-combining-selectors.test.ts
```diff
@@ -1,148 +1,153 @@
...
-ruleTester().run(path.parse(__filename).name, rule, {
-  valid: [
-    `
+const valid = [
+  `
 import { Store } from '@ngrx/store'
 
 class Ok {
-  readonly test$ = somethingOutside();
+readonly test$ = somethingOutside();
```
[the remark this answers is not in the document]
timdeschryver: Good catch, these were made by accident

tests/rules/avoid-duplicate-actions-in-reducer.test.ts
timdeschryver: Where would this be needed?

PR #260, opened 2021-10-13
fix: handle `ClassDeclaration` without `id` properly

rafaelss95:
**refactor: separate `TSClassImplements` to reuse**:

it extracts the `TSClassImplements` selector from the `prefer-action-creator` to `src/utils/selectors/` in order to reuse.

**refactor: report `ClassDeclaration[id]` instead of `MethodDefinition[key]`**:

IMHO, it makes more sense if we report `ClassDeclaration` instead of `MethodDefinition`, because from a technical point of view, we should indicate that the `class` should implement an `interface` and not that the `method` has "error".

**fix: handle `ClassDeclaration` without `id` properly**:

In fact, the fix has always been broken for `default class {`, since we had a typecast in the `id` (`id as TSETree.Identifier`) where the `id` could be in fact `null`.

### DIFF
--- a/src/rules/effects/use-effects-lifecycle-interface.ts
+++ b/src/rules/effects/use-effects-lifecycle-interface.ts
@@ -6,6 +6,7 @@ import {
   docsUrl,
   getImplementsSchemaFixer,
   getImportAddFix,
+  getLocOrNode,
   NGRX_MODULE_PATHS,
 } from '../../utils'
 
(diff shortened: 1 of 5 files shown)


timdeschryver (commented):
The reason that we added the lint to the method as to keep it in line with the angular lifecycle linter - https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/docs/rules/use-lifecycle-interface.md

Personally, I would want to keep the behavior consistent.
What do you think?

PR #262, opened 2021-10-21
fix(prefix-selectors-with-select): handle all kinds of IdentifierName properly

### LINKED ISSUE #261: Bug ngrx/prefix-selectors-with-select rule does not accept variables with numbers in their name
## Behavior

`ngrx/prefix-selectors-with-select` rule does not accept variables with numbers in their name.

## Code snippet

The following code
```ts
export const selectI18nFeature = createFeatureSelector<State>(i18nFeatureKey);
```
throws:
```
warning  The selector should start with "select"  ngrx/prefix-selectors-with-select
```

[ ] No

alorle:
As defined by ECMAScript spec (https://tc39.es/ecma262/#prod-IdentifierName), an identifier can contain
- `UnicodeIDContinue` (any Unicode code point with the Unicode property _ID_Continue_),
- `$`,
- `U+200C` (ZERO WIDTH NON-JOINER, \<ZWNJ\>),
- `U+200D` (ZERO WIDTH JOINER, \<ZWJ\>)
- `\ UnicodeEscapeSequence`

and must start with one of
- `UnicodeIDStart` (any Unicode code point with the Unicode property _ID_Start_),
- `$`,
- `_`

Due to the complexity of this specification and because I believe that this rule should not assume the style of the code it corrects (if using camelCase or snake_case, for example), I have removed the test `export const selectfeature = createSelector ((state: AppState) => state.feature)`. As I say, in my opinion this line of code should not fail, at least with this specific rule.

In addition to this, a test is added with the case of failure that has motivated this RP.

I remain at the disposal of any question.

PS: this is my first contribution to public software :smile:, don't lynch me :confused:...

Resolves #261

### DIFF
--- a/tests/rules/prefix-selectors-with-select.test.ts
+++ b/tests/rules/prefix-selectors-with-select.test.ts
@@ -31,6 +31,8 @@ const valid: RunTests['valid'] = [
     export const selectThing = (id: string) => createSelector(selectThings, things => things[id])`,
   `
     export const selectFeature = createSelectorFactory(factoryFn)`,
+  `
+    export const selectFeature1 = createSelectorFactory(factoryFn)`,
 ]
 
 const invalid: RunTests['invalid'] = [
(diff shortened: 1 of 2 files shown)


tests/rules/prefix-selectors-with-select.test.ts
```diff
@@ -124,24 +126,6 @@ const invalid: RunTests['invalid'] = [
       ],
     },
   ),
-  fromFixture(
-    stripIndent`
-      export const selectfeature = createSelector((state: AppState) => state.feature)
```
timdeschryver, line 129: We really want to have the prefix `select` followed by a domain.

I haven't seen snake_case been used in the wild, and this is also something I haven't thought of.
For me, it's also OK to use that as a selector name, so `select_feature` would also a valid name.
But I see `selectfeature` still as an invalid name, which should become `selelctFeature` or `select_feature`.
Later, we could also potentially add a config flag for this, e.g. `all | camelCase | snake_case`.

Thoughts?

alorle:
> That's super to hear! I hope it was a smooth process to setup the project and contribute to it. You also don't have to worry about being lynched, these changes are great and I'm sure it will help future devs who use this plugin!

Smooth process. Definitely 👌

PR #274, opened 2021-12-09
Update on-function-explicit-return-type.ts

bryantlikes:
if there are no parameters then just append the `: State `  to the empty arrow function

### DIFF
--- a/src/rules/store/on-function-explicit-return-type.ts
+++ b/src/rules/store/on-function-explicit-return-type.ts
@@ -54,10 +54,16 @@ export default createRule<Options, MessageIds>({
 })
 
 function getFixes(
-  { params }: TSESTree.ArrowFunctionExpression,
+  node: TSESTree.ArrowFunctionExpression,
   sourceCode: Readonly<TSESLint.SourceCode>,
   fixer: TSESLint.RuleFixer,
 ) {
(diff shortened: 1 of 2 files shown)


src/rules/store/on-function-explicit-return-type.ts
```diff
@@ -54,10 +54,16 @@ export default createRule<Options, MessageIds>({
...
 
 function getFixes(
-  { params }: TSESTree.ArrowFunctionExpression,
+  node: TSESTree.ArrowFunctionExpression,
   sourceCode: Readonly<TSESLint.SourceCode>,
   fixer: TSESLint.RuleFixer,
 ) {
+  const { params } = node;
+    
+  if (params.length === 0) {
```
timdeschryver, line 63: Thanks for creating the PR!
The implementation looks good, but it would be awesome if we could also include a test case.
This can be the code snippet that's posted in the issue.

bryantlikes: Added test and found/fixed an issue

timdeschryver (approved):
Thanks @bryantlikes !

timdeschryver:
@all-contributors please add @bryantlikes  for code, tests, bug

PR #292, opened 2022-07-18
fix: 🐛 downgrade required typescript version

danymarques:
This will allow projects using Angular 12 and npm 8 to install dependencies without the --force flag.


timdeschryver (approved):
Thanks!

